PR #2, opened 2015-05-04
EZP-24304: Provide a way to customize field definition form views

lolautruche:
> https://jira.ez.no/browse/EZP-24304

Added `getFieldDefinitionEditConfig()` to FieldTypeFormMapperInterface.
Returned value is a hash which may contain:
* `template`: template where specific form fields defined in mapFieldDefinitionForm() will be displayed.
* `vars`: Hash of additional variables to pass to the template.

Provided template will be included during FieldDefinition rendering, in order to add potential additional fields.
Variables passed to the form are:
* `data`: FieldDefinitionData object
* `languageCode`: Language code as passed to the main form
* `contentTypeDraft`: The ContentTypeDraft object
* `form`: Current FieldDefinition form

Note that it is *not mandatory* to provide a template. If a FieldType doesn't need additional form fields (i.e. no validator or field settings), there is no need to define a template.

Also moved update_content_type.html.twig to ContentType subdirectory

### Follow-up
* [EZP-24322](https://jira.ez.no/browse/EZP-24322): Use template blocks instead of including templates.

### DIFF
--- a/bundle/Controller/TestController.php
+++ b/bundle/Controller/TestController.php
@@ -109,8 +109,12 @@ class TestController extends Controller
             return $this->redirectToRoute('contenttype/update', ['contentTypeId' => $contentTypeId, 'languageCode' => $languageCode]);
         }
 
-        return $this->render('EzSystemsRepositoryFormsBundle::update_content_type.html.twig', [
-            'form' => $form->createView()
+        return $this->render('EzSystemsRepositoryFormsBundle:ContentType:update_content_type.html.twig', [
+            'form' => $form->createView(),
+            'contentTypeName' => $contentTypeDraft->getName($languageCode),
+            'contentTypeDraft' => $contentTypeDraft,
+            'languageCode' => $languageCode,
+            'fieldTypeMapperRegistry' => $this->get('ezrepoforms.field_type_form_mapper.registry')
         ]);
     }
 }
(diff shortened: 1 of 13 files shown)


lolautruche:
Review ping @bdunogier @dpobel @andrerom @pspanja

pspanja:
I guess template will be exposed to configuration later on?
Also, I think the same could also be achieved through form fragment block override as described in http://symfony.com/doc/current/cookbook/form/form_customization.html.

lolautruche:
No, template is not meant to be exposed to configuration. This is entirely up to the mapper.
I considered block override, but this is a much more complicated thing (see our fields management).

lolautruche:
After thinking about it, I think you're right. We should use template fragments, but not as for form customization, much more like what we use with `ez_render_field()`. This would allow one template file per field type, with different blocks (for field view, field edit, field def view, field def edit).

This would moreover allow us to refactor this part a bit, including config.
However, this cannot be done quickly so I suggest to do this as a follow up.

What do you think @bdunogier @dpobel @pspanja @andrerom ?

pspanja:
Yeah, that is what I had in mind. I'm ok for a follow-up :)

pspanja:
+1

bdunogier:
+1 with the follow-up.

andrerom:
+1 with followup :+1:

lolautruche:
Thanks :-)

Here's the follow-up: https://jira.ez.no/browse/EZP-24322

PR #3, opened 2015-05-06
Fix EZP-24309: Implement validation for ContentType forms

lolautruche:
> https://jira.ez.no/browse/EZP-24309

Adds validation for ContentTypeData and FieldDefinitionData.

* [x] Add validation for ContentTypeData fields
* [x] Add validation for common FieldDefinitionData fields
* [x] Translate custom validation messages
* [x] Ensure inner FieldDefinitionData fields are validated in cascade when validating ContentTypeData
* [x] Add validation for FieldDefinitionData validator configuration
* [x] Add validation for FieldDefinitionData field settings configuration
* [x] Add validation for FieldDefinitionData default value when needed (should also be usable for content field Value validation)
* [x] Tests for validators and constraints

Follow ups:
* Add a target property/getter to `\eZ\Publish\SPI\FieldType\ValidationError` so that we can easily identify on which field validation errors occur (at field type level)

### DIFF
--- a/bundle/Controller/TestController.php
+++ b/bundle/Controller/TestController.php
@@ -61,6 +61,7 @@ class TestController extends Controller
         foreach ($contentTypeDraft->fieldDefinitions as $fieldDef) {
             $formData->addFieldDefinitionData(new FieldDefinitionData([
                 'fieldDefinition' => $fieldDef,
+                'contentTypeData' => $formData,
                 'identifier' => $fieldDef->identifier,
                 'names' => $fieldDef->getNames(),
                 'descriptions' => $fieldDef->getDescriptions(),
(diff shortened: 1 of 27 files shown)


bundle/Controller/TestController.php
```diff
@@ -61,6 +61,7 @@ public function updateContentTypeAction(Request $request, $contentTypeId, $langu
         foreach ($contentTypeDraft->fieldDefinitions as $fieldDef) {
             $formData->addFieldDefinitionData(new FieldDefinitionData([
                 'fieldDefinition' => $fieldDef,
+                'contentTypeData' => $formData,
```
bdunogier: why `$formData` data and not `$contentTypeData` ?

lolautruche: Seems legit :-)

bdunogier:
Looks good so far :-)

lolautruche:
Now ready for review

Ping @bdunogier @pspanja @andrerom

andrerom:
+1, but @bdunogier and @pspanja should add their input here :)

lolautruche:
Also ping @glye

bdunogier:
+1

PR #6, opened 2015-05-12
Made it possible to select any of the registered FieldTypes…

lolautruche:
...when adding a FieldDefinition

All fieldtypes can work as is. They just won't have configuration field settings / validation config / default value.


lolautruche:
review ping @bdunogier @andrerom @glye @pspanja

andrerom:
+1

PR #10, opened 2015-05-21
Fix EZP-24397: Implement FieldDefinition form mapper for ezinteger

glye:
Manual test. Requires this change in the fieldvalue converter to work: https://github.com/ezsystems/ezpublish-kernel/pull/1264

https://jira.ez.no/browse/EZP-24397

### DIFF
--- a/lib/FieldType/Mapper/IntegerFormMapper.php
+++ b/lib/FieldType/Mapper/IntegerFormMapper.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\FormInterface;
+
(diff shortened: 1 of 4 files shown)


glye:
The doc says it uses IntegerValueValidator: https://doc.ez.no/display/EZP/The+Integer+FieldType
Dunno yet why it fails. Does it receive empty string, or string "0" maybe?

glye:
Rebased after the ezfloat commit. Review ping @lolautruche

glye:
Added 'empty_data' => 0 to min/max values, otherwise form complains of invalid value when submitting. Unsure if this is correct: How can you submit the default max value? (which is false, i.e. not a valid integer)

lib/FieldType/Mapper/IntegerFormMapper.php
```diff
@@ -0,0 +1,38 @@
...
+use Symfony\Component\Form\FormInterface;
+
+class IntegerFormMapper implements FieldTypeFormMapperInterface
+{
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+        $fieldDefinitionForm
+            ->add(
+                'minValue', 'integer', [
+                    'empty_data' => 0,
```
lolautruche: What about negative values?

glye:
Reverted 'empty_data' change, added better fix in https://github.com/ezsystems/ezpublish-kernel/pull/1264

glye:
Rebased due to upstream id changes in translation file, no code change.

pspanja:
+1

glye:
@lolautruche About negative values, 0 is the default for minvalue, and setting 0 for empty_data afaik does not stop you from using negative values. Anyway, I removed it, so no longer relevant.

lolautruche:
+1

PR #12, opened 2015-05-21
Fix EZP-24383: Implement form action dispatcher

lolautruche:
> https://jira.ez.no/browse/EZP-24383

This PR introduces a simple mechanism to handle several actions in a single form.
Every time a form using the mechanism is submitted (and validated), a `FormActionEvent` is fired. This event contains the form, its underlying data, the clicked button name and the language code.

In the case of ContentType, event names are `contentType.update.<buttonName>`, which makes it very easy to attach a specific *form processor* (which is actually an event subscriber).

`FormDataMapperInterface` is also introduced to help converting repository value objects to the underlying form data. This will help avoiding doing this in the controller.


lolautruche:
Review ping @andrerom @bdunogier @dpobel @yannickroger @pspanja @glye

bundle/Controller/ContentTypeController.php
```diff
@@ -0,0 +1,74 @@
...
+use eZ\Bundle\EzPublishCoreBundle\Controller;
+use eZ\Publish\Core\Repository\Values\ContentType\ContentTypeCreateStruct;
+use EzSystems\RepositoryForms\Data\Mapper\ContentTypeDraftMapper;
+use EzSystems\RepositoryForms\Event\FormActionEvent;
+use EzSystems\RepositoryForms\Event\RepositoryFormEvents;
+use Symfony\Component\HttpFoundation\Request;
+
+class ContentTypeController extends Controller
+{
+    public function createContentTypeAction($contentTypeGroupId, $languageCode = null)
```
bdunogier: Could we use the ParamConverter thing for the ContentTypeGroup ?

glye:
+1

bundle/Controller/ContentTypeController.php
lolautruche: OutOfScopeException :smile: 

Follow-up story created: https://jira.ez.no/browse/EZP-24409

bundle/Controller/ContentTypeController.php
```diff
@@ -0,0 +1,77 @@
...
+
+        return $this->redirectToRoute(
+            'contenttype/update',
+            ['contentTypeId' => $contentTypeDraft->id, 'languageCode' => $languageCode]
+        );
+    }
+
+    public function updateContentTypeAction(Request $request, $contentTypeId, $languageCode = null)
+    {
+        $languageCode = $languageCode ?: $this->getConfigResolver()->getParameter('languages')[0];
```
bdunogier: No controller as a service, and params/services injection ?

bdunogier:
+1 besides the remarks

bundle/Controller/ContentTypeController.php
lolautruche: Well, we'd have a lot of dependencies for very low gain IMHO.
Services needed:
* Templating
* Router
* FormFactory
* EventDispatcher
* ContentTypeService
* FieldTypeFormMapperRegistry

+ of course language list as injected setting.

lolautruche:
I tried to the controller as a service and finally ended up to rewrite most helper methods from `\Symfony\Bundle\FrameworkBundle\Controller\Controller`. So doesn't sound like a good fit here.

PR #14, opened 2015-05-22
Fix EZP-24381: As a RepositoryForms user, I want to be able to publish a content type draft

lolautruche:
> https://jira.ez.no/browse/EZP-24381

This PR simply adds the possibility to *publish* a ContentTypeDraft.
It also:
* Makes it possible to create a ContentTypeDraft out of an existing ContentType if it doesn't exist yet.
* Fixes ContentType identifier uniqueness validator to allow edition of an existing ContentType.

## TODO
* [x] Fix UniqueContentTypeIdentifierValidatorTest

### DIFF
--- a/bundle/Resources/translations/ezrepoforms_content_type.en.xlf
+++ b/bundle/Resources/translations/ezrepoforms_content_type.en.xlf
@@ -62,6 +62,10 @@
                 <source>content_type.publish</source>
                 <target>OK</target>
             </trans-unit>
+            <trans-unit id="16">
+                <source>content_type.remove_field_definitions</source>
+                <target>Remove selected field definitions</target>
+            </trans-unit>
             <!-- Field definition -->
             <trans-unit id="101">
                 <source>field_definition.name</source>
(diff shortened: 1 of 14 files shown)


lolautruche:
Review ping @andrerom @bdunogier @yannickroger @glye @pspanja

andrerom:
+1 (even if I'm not eagerly fan of try catch for create / update logic)

bundle/Resources/translations/ezrepoforms_content_type.en.xlf
```diff
@@ -58,6 +58,10 @@
                 <source>content_type.edit_title</source>
                 <target>Edit &lt;%contentTypeName%&gt;</target>
             </trans-unit>
+            <trans-unit id="15">
+                <source>content_type.publish</source>
+                <target>OK</target>
```
andrerom: "Ok" might not be the most clear action for publishing Content Type, should we go for "Publish"? Any input on what button should say form @rolandbenedetti maybe?

lolautruche: I kept translation from legacy

lolautruche:
@andrerom I'm not a fan either but this is currently the only way to do with the API

andrerom:
> @andrerom I'm not a fan either but this is currently the only way to do with the API

Alternative would be that UI knows and exposes different actions based on if it will first need to create a draft (action that creates draft and redirects to edit draft) or edit existing one directly. But that would be complexity in UI code, so I'm fine with this.

glye:
+1

PR #16, opened 2015-05-26
EZP-24379: Integrate ContentType edition with PlatformUI

lolautruche:
> https://jira.ez.no/browse/EZP-24379

Prepares integration into PlatformUIBundle.
See https://github.com/ezsystems/PlatformUIBundle/pull/236


lib/Form/ActionDispatcher/AbstractActionDispatcher.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->dispatchAction($defaultActionEventName . ($actionName ? ".$actionName" : ''), $event);
+        $this->response = $event->getResponse();
+    }
+
+    /**
+     * Configures options to pass to the form action event.
+     *
+     * @param OptionsResolver $resolver
+     */
+    protected function configureOptions(OptionsResolver $resolver)
```
andrerom: should this be abstract or should we document why it does nothing by default?

andrerom:
besides question, +1

lib/Form/ActionDispatcher/AbstractActionDispatcher.php
lolautruche: I considered setting this method abstract, but we may not have options to resolve, hence default implementation.
But maybe it makes more sense to have it abstract, yes. This also allows implementation to not do anything.

lolautruche: Opinion on my comment @andrerom ?

andrerom: for me both works, but abstract is more explicit so maybe the cleaner approach.
but then add doc that it can optionally also do nothing :)

lolautruche: Fixed in 8ad109c41125dee512b482b33330bd5824804076

lolautruche:
Review ping @pspanja @bdunogier @yannickroger @glye

PR #18, opened 2015-05-28
Fix EZP-24389: Implement FieldDefinition form mapper for ezcountry

glye:
https://jira.ez.no/browse/EZP-24389

- [x] Read country data from ezpublish.fieldType.ezcountry.data and show it in the form
- [x] Save default selection
- [x] Test the transformer

### DIFF
--- a/lib/FieldType/Mapper/CountryFormMapper.php
+++ b/lib/FieldType/Mapper/CountryFormMapper.php
@@ -1,0 +1,76 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use eZ\Publish\Core\FieldType\Country;
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\DataTransformer\CountryValueTransformer;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\Extension\Core\ChoiceList\ChoiceList;
+use Symfony\Component\Form\FormInterface;
+
+class CountryFormMapper implements FieldTypeFormMapperInterface
+{
+    /**
+     * @var array Array of countries from ezpublish.fieldType.ezcountry.data
+     */
+    protected $countriesInfo;
+
+    /**
+     * @param array $countriesInfo
+     */
+    public function __construct(array $countriesInfo)
+    {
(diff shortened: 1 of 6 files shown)


glye:
Review ping @lolautruche @pspanja @andrerom @bdunogier @yannickroger

andrerom:
So the FieldType does not support default selection is that what your saying?
We can create issue for that and handle it separately, as it might affect Platform UI as well ( ? @dpobel )

glye:
Yes, that's what I'm saying - if I'm reading it correctly. Am wondering if the feature may have been deprecated, or just forgotten.

lolautruche:
IMHO it has been forgotten. I guess it would not be very hard to implement this.

glye:
Rebased after ISBN mapper commit.

glye:
@lolautruche @andrerom Legacy reads the country data out of settings/country.ini:
https://github.com/ezsystems/ezpublish-legacy/blob/master/settings/country.ini
We can copy this to somewhere in new stack, or use the Symfony Intl component:
http://symfony.com/doc/current/components/intl.html
Intl holds only country name and 2-letter code. The old country.ini also holds a 3-letter code and phone dialing code, but those don't seem to be used in the ezcountry type. So I think Intl can be used here.

lolautruche:
We already have the data, and it's passed to the FT: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/settings/fieldtypes.yml#L28

glye:
Ah, there it is, thanks.

glye:
Now it reads country data from ezpublish.fieldType.ezcountry.data, but saving doesn't work yet.

lib/FieldType/Mapper/CountryFormMapper.php
```diff
@@ -0,0 +1,71 @@
...
+
+class CountryFormMapper implements FieldTypeFormMapperInterface
+{
+    /**
+     * @var array
+     */
+    protected $countriesInfo;
+
+    /**
+     * @param array $countriesInfo Array of countries from ezpublish.fieldType.ezcountry.data
```
lolautruche: Better move this to property doc instead

lib/FieldType/Mapper/CountryFormMapper.php
```diff
@@ -0,0 +1,71 @@
...
+                [
+                    'required' => false,
+                    'property_path' => 'fieldSettings[isMultiple]',
+                    'label' => 'field_definition.ezcountry.is_multiple',
+                ]
+            )
+            ->add(
+                'defaultSelection',
+                'choice',
+                [
```
lolautruche: CS: `[` should be on the same line than the last comma. Makes you save 1 indent

glye:
@lolautruche Fixed feedback on doc and CS, though phpcs complains that "Only one argument is allowed per line in a multi-line function call". Test is TODO.

glye:
Added tests, ready for review

lolautruche:
+1

pspanja:
+1

PR #19, opened 2015-06-02
Fix EZP-24394: Implement FieldDefinition form mapper for eztime

glye:
Manual testing. This is based on ezdatetime, but thankfully the tricky bits of that are not needed here, so it's pretty simple.

https://jira.ez.no/browse/EZP-24394

### DIFF
--- a/lib/FieldType/Mapper/TimeFormMapper.php
+++ b/lib/FieldType/Mapper/TimeFormMapper.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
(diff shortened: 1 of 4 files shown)


glye:
Fixed translation copypasta error.

pspanja:
+1

lib/FieldType/Mapper/TimeFormMapper.php
```diff
@@ -0,0 +1,46 @@
...
+                    'required' => false,
+                    'property_path' => 'fieldSettings[useSeconds]',
+                    'label' => 'field_definition.eztime.use_seconds',
+                ]
+            )
+            ->add(
+                'defaultType',
+                'choice',
+                [
+                    'choices' => array(
```
pspanja: (We should probably be consistent with short syntax.)

glye:
Short array syntax everywhere.

glye:
Rebased after ezcountry commit

lolautruche:
+1

bdunogier:
+1

PR #21, opened 2015-06-04
Fix EZP-24449: Implement FieldDefinition form mapper for RichText

glye:
Manual testing.

Also: Tag preset field setting removed in kernel: https://github.com/ezsystems/ezpublish-kernel/pull/1301

https://jira.ez.no/browse/EZP-24449

### DIFF
--- a/lib/FieldType/Mapper/RichTextFormMapper.php
+++ b/lib/FieldType/Mapper/RichTextFormMapper.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use eZ\Publish\Core\FieldType\RichText\Type;
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\FormInterface;
+
+class RichTextFormMapper implements FieldTypeFormMapperInterface
+{
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+        $fieldDefinitionForm
+            ->add('numRows', 'integer', [
+                'empty_data' => 10,
+                'required' => true,
+                'property_path' => 'fieldSettings[numRows]',
+                'label' => 'field_definition.ezrichtext.num_rows',
+            ]);
+    }
+}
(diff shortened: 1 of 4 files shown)


lib/FieldType/Mapper/RichTextFormMapper.php
```diff
@@ -0,0 +1,47 @@
...
+                'integer',
+                [
+                    'empty_data' => 10,
+                    'required' => true,
+                    'property_path' => 'fieldSettings[numRows]',
+                    'label' => 'field_definition.ezrichtext.num_rows',
+                ]
+            )
+            ->add(
+                'tagPreset',
```
andrerom: I'm not sure we want to re introduce this one in the same way, ping @dpobel @pspanja

pspanja: That setting is really a leftover from the time when RichText was improved XmlText field type.
It does not go deeper than UI, and as far as I'm concerned we can remove it. But maybe we should still ping PM about it.

dpobel: I agree it's more a PM decision, but in a first step, I would not add this settings because given the new behaviour of the RichText editor, the setting would most likely to be a bit different.

andrerom: @rolandbenedetti Ok if we skip this one for now? as @dpobel says we probably want to do this more inline with what is planned for UI here and also long time issue is that this should rather be permission handled\* if possible 


\* *However that opens a range of new issues again, like user editing existing RichText with tags he does not have access to add himself, how to handle that.. so maybe not.*

rolandbenedetti73: well to give an answer, I'd like to understand better what it really is, or intend to be. 
if like @pspanja says it does nothing, I don't see why we would keep it, but I'd like to understand better what it was designed for. I should probably know :-( but I don't...

andrerom: It was added to provide pre sets of buttons in ezoe/TinyMCE, it was intended to be done more elegantly in the future which never happened, so should not be carried over to RichText like this, hence my vote to remove and rather re introduce later when we are ready for it.

rolandbenedetti73: Ok, we discussed all together, it makes no sens in regards to the target UX.
We can skip this preset.
+1

glye: Thanks all! I'll remove the tag preset.

glye:
Rebased on master, and removed tagPreset as discussed. Will also need to remove it from the Type and tests, but that's a tale for another PR.

lib/FieldType/Mapper/RichTextFormMapper.php
```diff
@@ -0,0 +1,33 @@
...
+            ->add(
+                'numRows',
+                'integer',
+                [
+                    'empty_data' => 10,
+                    'required' => true,
+                    'property_path' => 'fieldSettings[numRows]',
+                    'label' => 'field_definition.ezrichtext.num_rows',
+                ]
+            );
```
lolautruche: Lots of indents here, can be simplified :smiley: 

```php
$fieldDefinitionForm->add('numRows, 'integer', [
    'empty_data' => 10,
    'required' => true,,
    'property_path' => 'fieldSettings[numRows]',
    'label' => 'field_definition.ezrichtext.num_rows',
]);
```

glye:
Reduced indents

andrerom:
+1

pspanja:
+1

glye:
Rebased on master

PR #22, opened 2015-06-05
Fix EZP-24441: Implement FieldDefinition form mapper for ezbinaryfile

glye:
Manual tests.

Depends on changes in Type and Converter, see: https://github.com/ezsystems/ezpublish-kernel/pull/1283

https://jira.ez.no/browse/EZP-24441

### DIFF
--- a/lib/FieldType/Mapper/BinaryFileFormMapper.php
+++ b/lib/FieldType/Mapper/BinaryFileFormMapper.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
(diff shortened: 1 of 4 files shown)


andrerom:
+1

lib/FieldType/Mapper/BinaryFileFormMapper.php
```diff
@@ -0,0 +1,31 @@
...
+        $fieldDefinitionForm
+            ->add(
+                'maxSize',
+                'integer',
+                [
+                    'required' => true,
+                    'property_path' => 'validatorConfiguration[FileSizeValidator][maxFileSize]',
+                    'label' => 'field_definition.ezbinaryfile.max_file_size',
+                ]
+            );
```
lolautruche: CS. PSR-2 suggests the following:

```php
$fieldDefinitionForm->add('maxSize', 'integer', [
    'required' => true,
    'property_path' => 'validatorConfiguration[FileSizeValidator][maxFileSize]',
    'label' => 'field_definition.ezbinaryfile.max_file_size',
]);
```

You gain one indent :wink:

lolautruche:
Except CS remark, +1

lib/FieldType/Mapper/BinaryFileFormMapper.php
glye: @lolautruche Interesting. PhpStorm says "Only one argument is allowed per line in a multi-line function call", though not if only the array declaration is multi-line. Ok, I'll do this.

glye:
CS remark fixed.

PR #23, opened 2015-06-09
Fix EZP-24444: Implement FieldDefinition form mapper for ezpage

glye:
Uses PageService to read available layouts, to populate select dropdown.

Depends on https://github.com/ezsystems/ezpublish-kernel/pull/1289

https://jira.ez.no/browse/EZP-24444

### DIFF
--- a/lib/FieldType/Mapper/PageFormMapper.php
+++ b/lib/FieldType/Mapper/PageFormMapper.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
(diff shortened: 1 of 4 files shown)


lib/FieldType/Mapper/PageFormMapper.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * @var PageService Provides layout list used in form selector
+     */
+    protected $pageService;
+
+    /**
+     * @param PageService $pageService
+     */
+    public function __construct(PageService $pageService = null)
```
lolautruche: Why `null` by default?

lolautruche:
+1 besides remark

lib/FieldType/Mapper/PageFormMapper.php
glye: Remains from experimentation :) I'll remove it.

glye:
Removed useless null default param.

andrerom:
+1

glye:
Rebased on master

PR #25, opened 2015-06-10
Follow-up to EZP-24441: Implement FieldDefinition form mapper for ezbinaryfile

glye:
Max size should be an optional setting. Follow-up to https://jira.ez.no/browse/EZP-24441

Review ping @lolautruche


lolautruche:
+1

You can merge this straight away :wink:

PR #26, opened 2015-06-10
Fix EZP-24443: Implement FieldDefinition form mapper for ezmedia

glye:
Manual tests. Similar to ezbinaryfile and ezimage, but no change to the Type needed.

https://jira.ez.no/browse/EZP-24443


lolautruche:
+1

andrerom:
+1

PR #28, opened 2015-06-10
Fix EZP-24448: Implement FieldDefinition form mapper for ezselection

glye:
Manual test. Using the `collection` field type: http://symfony.com/doc/current/reference/forms/types/collection.html

https://jira.ez.no/browse/EZP-24448

### DIFF
--- a/lib/FieldType/Mapper/SelectionFormMapper.php
+++ b/lib/FieldType/Mapper/SelectionFormMapper.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\Extension\Core\ChoiceList\ChoiceList;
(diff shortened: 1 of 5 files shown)


glye:
Rebased on master.

glye:
Selection item add/remove works now, but the approach is not OK. The solution is in sight, though.

lib/FieldType/Mapper/SelectionFormMapper.php
```diff
@@ -0,0 +1,43 @@
...
+        $fieldDefinitionForm
+            ->add('isMultiple', 'checkbox', [
+                'required' => false,
+                'property_path' => 'fieldSettings[isMultiple]',
+                'label' => 'field_definition.ezselection.is_multiple',
+            ])
+            ->add('options', 'collection', [
+                'type' => 'text',
+                'options' => [
+                    'required' => false,
```
lolautruche: Can stay on one line :)

lib/FieldType/Mapper/SelectionFormMapper.php
```diff
@@ -0,0 +1,43 @@
...
+                ],
+                'allow_add' => true,
+                'allow_delete' => true,
+                'delete_empty' => true,
+                'prototype' => true,
+                'prototype_name' => mt_rand(),
```
glye: This is the bit that should be fixed. Ideally it should be an incrementing number based on the number of selection entries, but we don't know that at this point, hence the large random number. Anything unique works, a collision would cause trouble.

glye:
Reduced whitespace and added doc

glye:
Not using mt_rand() anymore, should be OK now.

bdunogier:
Beautiful :-)

andrerom:
+1

bdunogier:
+1

PR #31, opened 2015-06-12
Fix EZP-24446: Implement ezrelation FieldDefinition form mapper

glye:
It's unclear how the field for "Default selection item" should be set up. In legacy this lets you browse for a node, and stores a node ID: https://github.com/ezsystems/ezpublish-legacy/blob/master/design/standard/templates/class/datatype/edit/ezobjectrelation.tpl#L16

In the Type, the setting is a string, and the validator accepts strings:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/Relation/Type.php#L91

However, the Converter then stores this in dataInt2, so any actual string value is lost:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/RelationConverter.php#L72

The doc agrees it is a string, doesn't specify how it should be used, but says it is for future use:
https://doc.ez.no/display/EZP/The+Relation+FieldType

I have implemented it as a text field, with a UDW button to start the browsing action, and a todo notice to change it to hidden when the UDW is available (not yet, says @dpobel). Before then the text field can be used for manual testing.

https://jira.ez.no/browse/EZP-24446

### DIFF
--- a/bundle/Resources/views/ContentType/field_types.html.twig
+++ b/bundle/Resources/views/ContentType/field_types.html.twig
@@ -116,6 +116,23 @@
     </div>
 {% endblock %}
 
+{% block ezobjectrelation_field_definition_edit %}
+    <div class="ezobjectrelation-settings selection-root">
+        {{- form_label(form.selectionRoot) -}}
+        {{- form_errors(form.selectionRoot) -}}
+        {{- form_widget(form.selectionRoot) -}}
+
+        <button data-universaldiscovery-title="{{"field_definition.ezobjectrelation.selection_root_udw_title"|trans({}, "ezrepoforms_content_type")}}"
+                class="ez-button-tree pure-button ez-font-icon ez-button ez-relation-pick-root-button"
+                data-relation-root-input-selector="#{{form.selectionRoot.vars.id}}"
+                data-relation-selected-root-name-selector="#{{form.selectionRoot.vars.id}}-selected-root-name"
+                >{{"field_definition.ezobjectrelation.selection_root_udw_button"|trans({}, "ezrepoforms_content_type")}}</button>
+        <div id="{{form.selectionRoot.vars.id}}-selected-root-name">
+            {{ render( controller( "ez_content:viewLocation", {'locationId': form.selectionRoot.vars.value, 'viewType': '_platformui_content_name'} ) ) }}
+        </div>
+    </div>
+{% endblock %}
+
 {% block ezobjectrelationlist_field_definition_edit %}
     <div class="ezobjectrelationlist-settings selection-default-location">
         {{- form_label(form.selectionDefaultLocation) -}}
--- a/lib/FieldType/Mapper/RelationFormMapper.php
+++ b/lib/FieldType/Mapper/RelationFormMapper.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use eZ\Publish\Core\FieldType\Relation\Type;
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\FormInterface;
+
+class RelationFormMapper implements FieldTypeFormMapperInterface
+{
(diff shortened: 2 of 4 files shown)


andrerom:
@glye ignore the doc, the source of truth is legacy for existing features we want to keep, adapt the rest if it is done wrong. As for browsing, ask one of the UI guys how to use that.

glye:
(Rebased on master.) The selection root (start location for browsing for relation) is still a text field. I added a UDW button to start the browsing action, and a todo notice to change it to hidden when the UDW is available. Before that, the text field can be used for manual testing. Review ping @andrerom @lolautruche @pspanja

lolautruche:
For UDW wiget, you should ask @dpobel

glye:
Forgot to mention it: I have asked him, and implemented it as he suggested. The review ping hereby extends to you as well, @dpobel :)

bundle/Resources/views/ContentType/field_types.html.twig
```diff
@@ -88,6 +88,23 @@
...
+        {{- form_errors(form.selectionMethod) -}}
+        {{- form_widget(form.selectionMethod) -}}
+    </div>
+
+    <div class="ezrelation-settings selection-root">
+        {{- form_label(form.selectionRoot) -}}
+        {{- form_errors(form.selectionRoot) -}}
+        {{- form_widget(form.selectionRoot) -}}
+
+        <button data-universaldiscovery-title="Select a start location for browsing for relations"
```
bdunogier: Translations ?

glye:
@bdunogier D'oh. Translations added.

glye:
Open question voiced in demo: Does the "selection method" (browse or dropdown) setting make sense, when we start using UDW? Should I just remove it? Waiting for @dpobel to be back and comment on it.

The same applies to the RelationList type: https://github.com/ezsystems/repository-forms/pull/32

glye:
Rebased on master, and added CSS class and input selector. Ref @mhyndle

glye:
Removed selectionMethod property, as discussed this morning.
Review ping @mhyndle @dpobel @bdunogier @lolautruche et al.

mhyndle:
+1

lib/FieldType/Mapper/RelationFormMapper.php
```diff
@@ -0,0 +1,28 @@
...
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\FormInterface;
+
+class RelationFormMapper implements FieldTypeFormMapperInterface
+{
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+        $fieldDefinitionForm
+            ->add('selectionRoot', 'text', [ // TODO: Replace 'text' with 'hidden' when the UDW is ready to use
```
lolautruche: Please create a follow-up issue for this :-)

lolautruche:
+1

mhyndle:
As @dpobel proposed in https://github.com/ezsystems/PlatformUIBundle/pull/261#discussion_r33553489 can you add container for name of selected content? The best would be if this container (div?) have got id attribute.

bundle/Resources/views/ContentType/field_types.html.twig
```diff
@@ -116,6 +116,19 @@
     </div>
 {% endblock %}
 
+{% block ezobjectrelation_field_definition_edit %}
+    <div class="ezobjectrelation-settings selection-root">
+        {{- form_label(form.selectionRoot) -}}
+        {{- form_errors(form.selectionRoot) -}}
+        {{- form_widget(form.selectionRoot) -}}
+
+        <button data-universaldiscovery-title="{{"field_definition.ezobjectrelation.selection_root_udw_title"|trans({}, "ezrepoforms_content_type")}}"
```
mhyndle: Please add `data-relation-selected-root-name-selector` attribute with selector to container which display content's name (ref: https://github.com/ezsystems/repository-forms/pull/31#issuecomment-117188410)

glye:
@mhyndle @lolautruche Added attributes to the button, and a div with the content name, same as for relationlist. Added follow-up issue https://jira.ez.no/browse/EZP-24574
Again, I've got the +1's so I'll merge when Travis is happy.

bundle/Resources/views/ContentType/field_types.html.twig
glye: @mhyndle Done.

lib/FieldType/Mapper/RelationFormMapper.php
glye: @lolautruche Follow-up: https://jira.ez.no/browse/EZP-24574

mhyndle:
:thumbsup:

PR #32, opened 2015-06-12
Fix EZP-24447: Implement ezrelationlist FieldDefinition form mapper

glye:
Manual tests.

This is based on the Relation PR: https://github.com/ezsystems/repository-forms/pull/31
UDW button used in the same way.

https://jira.ez.no/browse/EZP-24447

### DIFF
--- a/bundle/Resources/config/kernel.yml
+++ b/bundle/Resources/config/kernel.yml
@@ -2,3 +2,8 @@ system:
     default:
         fielddefinition_edit_templates:
             - { template: "EzSystemsRepositoryFormsBundle:ContentType:field_types.html.twig", priority: 0 }
+        location_view:
+            _platformui_content_name:
+                content_name:
+                    template: "EzSystemsRepositoryFormsBundle:content_name:content_name.html.twig"
+                    match: {}
--- a/bundle/Resources/views/ContentType/field_types.html.twig
+++ b/bundle/Resources/views/ContentType/field_types.html.twig
@@ -116,6 +116,29 @@
     </div>
 {% endblock %}
 
+{% block ezobjectrelationlist_field_definition_edit %}
+    <div class="ezobjectrelationlist-settings selection-default-location">
+        {{- form_label(form.selectionDefaultLocation) -}}
+        {{- form_errors(form.selectionDefaultLocation) -}}
+        {{- form_widget(form.selectionDefaultLocation) -}}
+
+        <button data-universaldiscovery-title="{{"field_definition.ezobjectrelationlist.selection_root_udw_title"|trans({}, "ezrepoforms_content_type")}}"
+                class="ez-button-tree pure-button ez-font-icon ez-button ez-relation-pick-root-button"
+                data-relation-root-input-selector="#{{form.selectionDefaultLocation.vars.id}}"
+                data-relation-selected-root-name-selector="#{{form.selectionDefaultLocation.vars.id}}-selected-root-name"
+                >{{"field_definition.ezobjectrelationlist.selection_root_udw_button"|trans({}, "ezrepoforms_content_type")}}</button>
+        <div id="{{form.selectionDefaultLocation.vars.id}}-selected-root-name">
+            {{ render( controller( "ez_content:viewLocation", {'locationId': form.selectionDefaultLocation.vars.value, 'viewType': '_platformui_content_name'} ) ) }}
+        </div>
+    </div>
+
+    <div class="ezobjectrelationlist-settings selection-content-types">
+        {{- form_label(form.selectionContentTypes) -}}
+        {{- form_errors(form.selectionContentTypes) -}}
+        {{- form_widget(form.selectionContentTypes) -}}
+    </div>
+{% endblock %}
+
 {% block ezpage_field_definition_edit %}
     <div class="ezpage-settings default-layout">
         {{- form_label(form.defaultLayout) -}}
--- a/lib/FieldType/Mapper/RelationListFormMapper.php
+++ b/lib/FieldType/Mapper/RelationListFormMapper.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\Core\FieldType\RelationList\Type;
+use eZ\Publish\Core\Helper\TranslationHelper;
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\FormInterface;
+
+
+class RelationListFormMapper implements FieldTypeFormMapperInterface
+{
+    /**
+     * @var ContentTypeService Used to fetch list of available content types
+     */
+    protected $contentTypeService;
+
+    /**
+     * @var TranslationHelper Translation helper, for translated content type names
+     */
+    protected $translationHelper;
+
+    /**
+     * @param ContentTypeService $contentTypeService
+     * @param TranslationHelper $translationHelper
+     */
+    public function __construct(ContentTypeService $contentTypeService, TranslationHelper $translationHelper)
+    {
+        $this->contentTypeService = $contentTypeService;
+        $this->translationHelper = $translationHelper;
+    }
+
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+        // Fill a hash with all content types and their ids
+        $contentTypeHash = [];
+        foreach ($this->contentTypeService->loadContentTypeGroups() as $contentTypeGroup) {
+            foreach ($this->contentTypeService->loadContentTypes($contentTypeGroup) as $contentType) {
+                $contentTypeHash[$contentType->identifier] = $this->translationHelper->getTranslatedByProperty($contentType, "names");
+            }
+        }
+        sort($contentTypeHash);
+
+        $fieldDefinitionForm
+            ->add('selectionDefaultLocation', 'text', [ // TODO: Replace 'text' with 'hidden' when the UDW is ready to use
+                'required' => false,
+                'property_path' => 'fieldSettings[selectionDefaultLocation]',
+                'label' => 'field_definition.ezobjectrelationlist.selection_default_location',
+            ])
+            ->add('selectionContentTypes', 'choice', [
+                'choices' => $contentTypeHash,
+                'expanded' => false,
+                'multiple' => true,
+                'required' => false,
+                'property_path' => 'fieldSettings[selectionContentTypes]',
+                'label' => 'field_definition.ezobjectrelationlist.selection_content_types',
+            ]);
+    }
+}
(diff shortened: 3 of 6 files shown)


glye:
Rebased on master, and added the UDW button same as in https://github.com/ezsystems/repository-forms/pull/31

glye:
Review ping @dpobel @lolautruche @pspanja @bdunogier et al.

glye:
Open question voiced in demo: Does the "selection method" (browse or dropdown) setting make sense, when we start using UDW? Should I just remove it? CC @andrerom

The same applies to the Relation type: https://github.com/ezsystems/repository-forms/pull/31

andrerom:
I would say it does have a value, however it is just as well PM we should ask here.
Drop down of selections makes sense for when you use relations as a selection basically.

glye:
Rebased on master, added CSS class and input selector, and removed selectionMethod property, as discussed this morning.

mhyndle:
+1

lib/FieldType/Mapper/RelationListFormMapper.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    protected $configResolver;
+
+    /**
+     * @param Repository $repository
+     * @param ConfigResolverInterface $configResolver
+     */
+    public function __construct(Repository $repository, ConfigResolverInterface $configResolver)
+    {
+        $this->repository = $repository;
```
lolautruche: You don't need the whole repository here, only the ContentService. Please only inject it.

lib/FieldType/Mapper/RelationListFormMapper.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+        // Fill a hash with all content types and their ids
+        $contentTypeHash = [];
+        $contentTypeService = $this->repository->getContentTypeService();
+        foreach ($contentTypeService->loadContentTypeGroups() as $contentTypeGroup) {
+            foreach ($contentTypeService->loadContentTypes($contentTypeGroup) as $contentType) {
+                $contentTypeHash[$contentType->identifier] = $contentType->getName(
+                    $this->configResolver->getParameter("languages")[0]
```
lolautruche: Use translation helper instead. It's the service behind `ez_content_name()` Twig helper.

lolautruche: Furthermore, a getParameter() call in a 2 levels loop? :wink:

bundle/Resources/views/ContentType/field_types.html.twig
```diff
@@ -116,6 +116,25 @@
     </div>
 {% endblock %}
 
+{% block ezobjectrelationlist_field_definition_edit %}
+    <div class="ezobjectrelationlist-settings selection-default-location">
+        {{- form_label(form.selectionDefaultLocation) -}}
+        {{- form_errors(form.selectionDefaultLocation) -}}
+        {{- form_widget(form.selectionDefaultLocation) -}}
+
+        <button data-universaldiscovery-title="{{"field_definition.ezobjectrelationlist.selection_root_udw_title"|trans({}, "ezrepoforms_content_type")}}"
```
mhyndle: The same as for ezrelation (https://github.com/ezsystems/repository-forms/pull/31#discussion_r33573600):
* please add container for displaying selected content's name
* please add `data-relation-selected-root-name-selector` attribute in button which will contain selector pointing on container with selected content's name

glye:
@lolautruche improvements: Use ContentTypeService rather than repository, use TranslationHelper to translate content type names. This requires also https://github.com/ezsystems/ezpublish-kernel/pull/1324

glye:
@mhyndle Added container for displaying selected content's name, and data-relation-selected-root-name-selector attribute in button.

bundle/Resources/views/ContentType/field_types.html.twig
glye: Done

lib/FieldType/Mapper/RelationListFormMapper.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->translationHelper = $translationHelper;
+    }
+
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+        // Fill a hash with all content types and their ids
+        $contentTypeHash = [];
+        foreach ($this->contentTypeService->loadContentTypeGroups() as $contentTypeGroup) {
+            foreach ($this->contentTypeService->loadContentTypes($contentTypeGroup) as $contentType) {
+                $contentTypeHash[$contentType->identifier] = $this->translationHelper->getTranslatedContentTypeName($contentType);
```
lolautruche: Better use `$this->translationHelper->getTranslatedByProperty()`. It avoids to implement a new method for the helper.

bundle/Resources/views/ContentType/field_types.html.twig
```diff
@@ -116,6 +116,27 @@
...
+                class="ez-button-tree pure-button ez-font-icon ez-button ez-relation-pick-root-button"
+                data-relation-root-input-selector="#{{form.selectionDefaultLocation.vars.id}}"
+                data-relation-selected-root-name-selector="#selected_item_name"
+                >{{"field_definition.ezobjectrelationlist.selection_root_udw_button"|trans({}, "ezrepoforms_content_type")}}</button>
+        <div id="selected_item_name"></div>
```
mhyndle: I suppose that `id` attribute should be generated dynamically as there may be many fields that type in a contenttype. Then this ID should be passed to `data-relation-selected-root-name-selector` above.
What's more this `<div></div>` should as default contain the name of already selected content.

dpobel: the `id` won't be unique in the page if there are several relation list field definitions in a content type so it should contain the field definition identifier or something like that. In addition, you should use `-` as a separator instead of `_` and we can probably find a better name like `relation-selected-root-name` to be consistent with the button data attribute. And last, this should be pre-filled with the name of content if the field definition already has a selected root.

glye:
@lolautruche Using getTranslatedByProperty() for the content type name.

lib/FieldType/Mapper/RelationListFormMapper.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->translationHelper = $translationHelper;
+    }
+
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+        // Fill a hash with all content types and their ids
+        $contentTypeHash = [];
+        foreach ($this->contentTypeService->loadContentTypeGroups() as $contentTypeGroup) {
+            foreach ($this->contentTypeService->loadContentTypes($contentTypeGroup) as $contentType) {
+                $contentTypeHash[$contentType->identifier] = $this->translationHelper->getTranslatedByProperty($contentType,"names");
```
lolautruche: CS: missing WS

lolautruche:
+1

glye:
@mhyndle @dpobel Made the id unique by prefixing the field definition identifier. Added line view of the selected location, if set. Not at all sure if "ez_content:viewLocation", "viewType: line" is kosher here...

glye:
@lolautruche Fixed whitespace snafu.

glye:
@mhyndle I know you have +1-ed already, but given the changes above, can you please verify the approach? If it is okay, I'll do the same in https://github.com/ezsystems/repository-forms/pull/31

mhyndle:
The part which will be used in PlatformUI is OK for me
+1

glye:
@lolautruche Would it be insane to make a custom viewType that outputs only the content name? (I haven't found a pure twig way to get content name from location id - that would be the easiest if possible.)

lolautruche:
> I haven't found a pure twig way to get content name from location id

From locationId, it's obviously not possible. However, it is from location object:

```twig
{{ ez_content_name(location.contentInfo) }}
```

Otherwise, yeah, feel free to create a new viewType

glye:
Added new viewType for showing name of selected root node

bundle/Resources/config/kernel.yml
```diff
@@ -2,3 +2,9 @@ system:
     default:
         fielddefinition_edit_templates:
             - { template: "EzSystemsRepositoryFormsBundle:ContentType:field_types.html.twig", priority: 0 }
+        location_view:
+            content_name:
+                landing_page:
+                    template: "EzSystemsRepositoryFormsBundle:content_name:content_name.html.twig"
+                    match:
+                        Identifier\ContentType: [landing_page,folder]
```
glye: Problem: I see no way to match all/any. I don't care which content type the linked location refers to. But the yml demands a match section. Is there a way to use template X when viewType Y is used, no matter what content we're looking at?

lolautruche: AFAIK, just don't include `match`. That should work (unsure)

bundle/Resources/views/ContentType/field_types.html.twig
```diff
@@ -116,6 +116,29 @@
...
+        {{- form_errors(form.selectionDefaultLocation) -}}
+        {{- form_widget(form.selectionDefaultLocation) -}}
+
+        <button data-universaldiscovery-title="{{"field_definition.ezobjectrelationlist.selection_root_udw_title"|trans({}, "ezrepoforms_content_type")}}"
+                class="ez-button-tree pure-button ez-font-icon ez-button ez-relation-pick-root-button"
+                data-relation-root-input-selector="#{{form.selectionDefaultLocation.vars.id}}"
+                data-relation-selected-root-name-selector="#{{form.selectionDefaultLocation.vars.id}}-selected-root-name"
+                >{{"field_definition.ezobjectrelationlist.selection_root_udw_button"|trans({}, "ezrepoforms_content_type")}}</button>
+        <div id="{{form.selectionDefaultLocation.vars.id}}-selected-root-name">
+            {{ render_esi( controller( "ez_content:viewLocation", {'locationId': form.selectionDefaultLocation.vars.value, 'viewType': 'content_name'} ) ) }}
```
lolautruche: Why rendering an ESI here?

bundle/Resources/config/kernel.yml
glye: No, the controller demands a match in my test.

bundle/Resources/views/ContentType/field_types.html.twig
glye: Because copypasta :)

bundle/Resources/config/kernel.yml
lolautruche: Please try the following:

```yaml
default:
    location_view:
        content_name:
            default_view: 
                template: "EzSystemsRepositoryFormsBundle:content_name:content_name.html.twig"
                match: {}                
```

Furthermore, we should try to use a more specific viewType, e.g. `_platformui_content_name`, to be sure we avoid name collisions.

lolautruche: For information, matching is done here: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/MVC/Symfony/Matcher/AbstractMatcherFactory.php#L103

glye:
@lolautruche match {} works, thanks! Also changed the viewType as suggested. So the PR as a whole seems OK to me now.

bundle/Resources/config/kernel.yml
```diff
@@ -2,3 +2,8 @@ system:
     default:
         fielddefinition_edit_templates:
             - { template: "EzSystemsRepositoryFormsBundle:ContentType:field_types.html.twig", priority: 0 }
+        location_view:
+            _platformui_content_name:
+                content_name:
```
lolautruche: `default_view` looks more logical to me

glye: @lolautruche ...but default_view doesn't work (InvalidConfigurationException, Unrecognized option "default_view" under "ezpublish.system.default") and since I'm using `ez_content:viewLocation`, it makes sense to me :)

glye:
@mhyndle @lolautruche Travis stopped complaining, and I've already got +1's from you, so I'll merge soon unless I hear any protests on the latest changes, KTNXBYE.

mhyndle:
confirm my +1 ;)

PR #33, opened 2015-06-24
Fix EZP-24445: Add confirmation messages for ContentType edition

lolautruche:
> https://jira.ez.no/browse/EZP-24445

Title says it all...

Related to https://github.com/ezsystems/PlatformUIBundle/pull/260

### DIFF
--- a/lib/Form/Processor/ContentTypeFormProcessor.php
+++ b/lib/Form/Processor/ContentTypeFormProcessor.php
@@ -15,7 +15,9 @@ use EzSystems\RepositoryForms\Event\FormActionEvent;
 use EzSystems\RepositoryForms\Event\RepositoryFormEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Translation\TranslatorInterface;
 
 class ContentTypeFormProcessor implements EventSubscriberInterface
 {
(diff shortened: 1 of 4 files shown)


lolautruche:
Review ping @bdunogier @andrerom @glye @pspanja

lib/Form/Processor/ContentTypeFormProcessor.php
```diff
@@ -116,5 +133,20 @@ public function processRemoveContentTypeDraft(FormActionEvent $event)
...
+
+        $this->addNotification('content_type.notification.draft_removed');
+    }
+
+    /**
+     * Pushes a flash notification to session.
+     *
+     * @param string $message
+     */
+    private function addNotification($message, array $params = [])
```
bdunogier: Isn't this something we would move to a service somehow ? A `NotificationFlashBag` thing, that receives the session, and expects the notification string ?

lolautruche: I don't really see the need for a service here since the session is already a service we use for that.
As for the translator I'd say the same. Furthermore, the *context* (here `ezrepoforms_content_type`) will change, so it won't be that generic.

That's mainly why I chose a private method over a dedicated service.

bdunogier:
Will you update this PR according to what was discussed this morning (Notification value object, with a NotificationMessage that has a Translatable implementation) ?

lolautruche:
Yes, it will be closed.

Le ven. 26 juin 2015 15:21, Bertrand Dunogier <notifications@github.com> a
écrit :

lolautruche:
Closing as it's no longer needed given state of https://github.com/ezsystems/PlatformUIBundle/pull/260

PR #34, opened 2015-06-30
[req] Changed branch-alias to 0.2.x-dev

bdunogier:
As we're gonna tag 0.1.0.
Shouldn't require changes to ezplatform, as it requires repository-forms: ~0.1@dev, meaning that it should pick the branch alias instead, right ?


lolautruche:
+1

PR #35, opened 2015-07-08
Fix EZP-24577: Cannot add new relation/relationlist field types to content type

glye:
Added check for empty value. D'oh!

https://jira.ez.no/browse/EZP-24577


glye:
Review ping @mhyndle @lolautruche

lolautruche:
+1

mhyndle:
+1

PR #41, opened 2015-08-25
Fix EZP-24701: Role versioning

glye:
Makes use of role draft features added in main PR in ezpublish-kernel.

I removed the RoleFormProcessor I had here earlier. I moved code from here to the RoleFormProcessor in PlatformUI because I needed access to the notification feature it has, and I thought repository-forms should not depend on PlatformUI. On the other hand it means role editing will only work in PlatformUI. Unsure if this is the right way.

Requirement for https://github.com/ezsystems/ezpublish-kernel/pull/1376

https://jira.ez.no/browse/EZP-24701

### DIFF
--- a/lib/Form/Processor/RoleFormProcessor.php
+++ b/lib/Form/Processor/RoleFormProcessor.php
@@ -9,6 +9,7 @@
 namespace EzSystems\RepositoryForms\Form\Processor;
 
 use eZ\Publish\API\Repository\RoleService;
+use eZ\Publish\API\Repository\Values\User\RoleDraft;
 use EzSystems\RepositoryForms\Event\FormActionEvent;
 use EzSystems\RepositoryForms\Event\RepositoryFormEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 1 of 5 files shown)


glye:
Moved logic from platformui to here, leaving only notification and response in platformui.

glye:
Travis seems broken: "Array to string conversion - The command "phpunit --coverage-text" exited with 2."

lolautruche:
@glye Did you try locally?

glye:
@lolautruche Didn't get that far, my VM doesn't have enough memory to install the composer bits needed for the test (this works for kernel). Will try again when I have fixed the failing tests in kernel.

glye:
Made an insignificant change, to restart Travis.

lib/Form/Processor/RoleFormProcessor.php
```diff
@@ -36,24 +37,23 @@ public static function getSubscribedEvents()
...
     }
 
     public function processSaveRole(FormActionEvent $event)
     {
-        /** @var \EzSystems\RepositoryForms\Data\RoleData $roleData */
-        $roleData = $event->getData();
-        $this->roleService->updateRole($roleData->role, $roleData);
+        /** @var RoleDraft $roleDraft */
+        $roleDraft = $event->getData()->roleDraft;
+        $this->roleService->publishRoleDraft($roleDraft);
```
pspanja: Shouldn't publish be in a separate process, similar to ContentType?

glye: @pspanja It is in a separate process...? I mean, saving (updateRoleDraft) is done in `processDefaultAction`, and publishing in `processSaveRole`. Same as for content type afaict, other than the name of the method. I could rename `processSaveRole` to `processPublishRole`...?

pspanja: You're right, ignore it :)

pspanja:
+1

glye:
Rebased on master

lolautruche:
+1, but you should add test coverage on the mapper.

glye:
Added role mapper test.

PR #42, opened 2015-08-26
EZP-24727: Add possibility to inject class for div groups

andrerom:
https://jira.ez.no/browse/EZP-24727

Ready for review


andrerom:
ping @glye and @yannickroger

yannickroger:
+1

glye:
+1

PR #43, opened 2015-08-31
Small fixes in NewnessChecker & ContentTypeDraftMapper

lolautruche:
* Changed getIdentifier() method name + visibility in NewnessChecker trait, to ensure no collision with regular public getters from value objects
* Fixed ContentTypeDraftMapper. All properties except identifier must be copied from ContentTypeDraft
when a new one is created


lolautruche:
Review ping @andrerom @glye @bdunogier @yannickroger @pspanja

andrerom:
+1

glye:
+1

PR #44, opened 2015-09-07
Fix EZP-24752: Cancel buttons should be not trigger validation in PlatformUI

lolautruche:
> https://jira.ez.no/browse/EZP-24752

This patch deactivates validation when submitting ContentType / Role editing forms with the cancel button.
For this to completely work, a piece of JS should be implemented in PlatformUI to deactivate HTML5 validation as well.


lolautruche:
Review ping @andrerom @bdunogier @glye @dpobel @yannickroger @pspanja

glye:
+1

dpobel:
+1

PR #47, opened 2015-09-29
Fix: Service ez.repoforms.action_dispatcher.section not found exception

asfram:
The service 'ezsystems.platformui.controller.section' has a dependency on a non-existent service 'ezrepoforms.action_dispatcher.section'


andrerom:
Thanks, LGTM, but I'll let @lolautruche and @glye be the judge of that, maybe definition is still in PlatformUI repo, and hence need a bit more cleanup.

lolautruche:
Hi

Sorry but all section related services are [correctly declared in `section.yml`](https://github.com/ezsystems/repository-forms/blob/master/bundle/Resources/config/section.yml). And [`section.yml` is imported in main `services.yml`](https://github.com/ezsystems/repository-forms/blob/master/bundle/Resources/config/services.yml#L1-L2).

Ensure to always be up-to-date when testing PlatformUI + RepositoryForms :wink:

PR #50, opened 2015-10-19
Fix EZP-24713: Policy limitations edition

lolautruche:
> https://jira.ez.no/browse/EZP-24713

Works on the same principe as FieldDefinitions. Each limitation type needs a form mapper.
Base implementation includes a mapper for SiteAccessLimitationType

## TODO
* [x] Template mechanism
* [x] SiteAccessLimitationMapper
* [x] ContentTypeLimitationMapper
* [x] SectionLimitationMapper
* [x] Tests

### DIFF
--- a/bundle/DependencyInjection/Compiler/LimitationFormMapperPass.php
+++ b/bundle/DependencyInjection/Compiler/LimitationFormMapperPass.php
@@ -1,0 +1,43 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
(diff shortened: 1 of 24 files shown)


andrerom:
cool

> Each limitation type needs a form mapper.

Could you expand a bit on that? Makes kind of sense as we have similar things for Field Types in Content Types *(and probably Content)*.

lolautruche:
@andrerom Well, each limitation type needs a form to be properly edited. Most can just rely on the `MultipleSelectionBasedMapper`, but it is still needed to provide valid selectable values.

lolautruche:
Now ready for review
Ping @andrerom @bdunogier @pspanja @glye

andrerom:
+1

glye:
+1 afaict

PR #51, opened 2015-10-21
EZP-22658: Add Repo form support for Languages

andrerom:
https://jira.ez.no/browse/EZP-22658

### DIFF
--- a/lib/Data/Mapper/LanguageMapper.php
+++ b/lib/Data/Mapper/LanguageMapper.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Data\Mapper;
+
+use eZ\Publish\API\Repository\Values\Content\Language;
+use eZ\Publish\API\Repository\Values\ValueObject;
+use EzSystems\RepositoryForms\Data\Language\LanguageCreateData;
+use EzSystems\RepositoryForms\Data\Language\LanguageUpdateData;
+
+class LanguageMapper implements FormDataMapperInterface
+{
+    /**
+     * Maps a ValueObject from eZ content repository to a data usable as underlying form data (e.g. create/update struct).
+     *
+     * @param ValueObject|\eZ\Publish\API\Repository\Values\Content\Language $language
+     * @param array $params
+     *
+     * @return LanguageCreateData|LanguageUpdateData
+     */
+    public function mapToFormData(ValueObject $language, array $params = [])
+    {
+        if (!$this->isLanguageNew($language)) {
+            $data = new LanguageUpdateData(['language' => $language]);
+            $data->languageCode = $language->languageCode;
+            $data->name = $language->name;
+            $data->enabled = $language->enabled;
+        } else {
+            $data = new LanguageCreateData(['language' => $language]);
+        }
+
+        return $data;
+    }
+
+    private function isLanguageNew(Language $language)
+    {
+        return $language->id === null;
+    }
+}
--- a/lib/Event/RepositoryFormEvents.php
+++ b/lib/Event/RepositoryFormEvents.php
@@ -76,4 +76,9 @@ final class RepositoryFormEvents
      * Triggered when updating a section.
      */
     const SECTION_UPDATE = 'section.update';
+
+    /**
+     * Triggered when updating a language.
+     */
+    const LANGUAGE_UPDATE = 'language.update';
 }
--- a/lib/Form/Type/Language/LanguageDeleteType.php
+++ b/lib/Form/Type/Language/LanguageDeleteType.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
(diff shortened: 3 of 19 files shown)


lib/Data/Mapper/LanguageMapper.php
```diff
@@ -0,0 +1,43 @@
...
+
+use eZ\Publish\API\Repository\Values\Content\Language;
+use eZ\Publish\API\Repository\Values\ValueObject;
+use EzSystems\RepositoryForms\Data\Language\LanguageCreateData;
+use EzSystems\RepositoryForms\Data\Language\LanguageUpdateData;
+
+class LanguageMapper implements FormDataMapperInterface
+{
+    /**
+     * Maps a ValueObject from eZ content repository to a data usable as underlying form data (e.g. create/update struct).
```
bdunogier: I overall wouldn't repeat the interface's / parent's phpdoc unless it adds value.

lib/Validator/Constraints/UniqueLanguageCodeValidator.php
```diff
@@ -0,0 +1,61 @@
...
+     * Checks if the passed value is valid.
+     *
+     * @param LanguageUpdateData $value The value that should be validated
+     * @param Constraint|UniqueFieldDefinitionIdentifier $constraint The constraint for the validation
+     *
+     * @api
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        // REVIEW NOTE: Should this constraint not be checked on creation also? (if so, same with section)
```
bdunogier: Don't really see why it wouldn't be checked on creation as well.

lib/Data/Mapper/LanguageMapper.php
andrerom: ok, it's copy paste from Section, but can strip it.

lib/Validator/Constraints/UniqueLanguageCodeValidator.php
andrerom: I guess it was checked on both because of the namespace bug I fixed in 8ae8008, but @lolautruche might know for sure.

lib/Data/Mapper/LanguageMapper.php
lolautruche: We repeat it because of phpdoc param types

lib/Event/RepositoryFormEvents.php
```diff
@@ -76,4 +76,9 @@
      * Triggered when updating a section.
      */
     const SECTION_UPDATE = 'section.update';
+
+    /**
+     * Triggered when updating a section.
```
lolautruche: s/section/language

lib/Form/Type/Language/LanguageDeleteType.php
```diff
@@ -0,0 +1,44 @@
...
+class LanguageDeleteType extends AbstractType
+{
+    /**
+     * @var LanguageService
+     */
+    private $languageService;
+
+    public function __construct(LanguageService $languageService)
+    {
+        $this->languageService = $languageService;
```
lolautruche: You don't seem to need the language service here

lib/Form/Type/Language/LanguageType.php
```diff
@@ -0,0 +1,37 @@
...
+
+    public function getName()
+    {
+        return 'ezrepoforms_language_edit';
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'data_class' => '\eZ\Publish\API\Repository\Values\Content\LanguageStruct',
```
lolautruche: `LanguageStruct` doesn't exist... As far as I can see, you can set this to `LanguageCreateStruct`

lib/Validator/Constraints/UniqueLanguageCodeValidator.php
```diff
@@ -0,0 +1,61 @@
...
+namespace EzSystems\RepositoryForms\Validator\Constraints;
+
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\LanguageService;
+use EzSystems\RepositoryForms\Data\Language\LanguageUpdateData;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Will check if ContentType identifier is not already used in the content repository.
```
lolautruche: Wrong comment

lib/Form/Type/Language/LanguageDeleteType.php
andrerom: true

lib/Validator/Constraints/UniqueLanguageCodeValidator.php
lolautruche: +1 with @bdunogier

lib/Form/Type/Language/LanguageType.php
andrerom: true

lib/Validator/Constraints/UniqueLanguageCodeValidator.php
lolautruche: @andrerom You're right, there is an issue with Section. Should check the trait type (same here).

andrerom: Can't really check instanceof trait, so changed to check instance of API struct, alternative is to check if one of the data classes.

glye:
+1

PR #52, opened 2015-10-23
Fix EZP-25026: Content type identifier validation message should say latin1

glye:
Fixed message translations to say a-z, and made them more similar.

https://jira.ez.no/browse/EZP-25026


andrerom:
+1

bdunogier:
+1

PR #53, opened 2015-10-23
Fix EZP-24942: Press <Enter> in repository form raises a notification error

glye:
The default form action is always dispatched first. When no action is specified (as when pressing enter in a repoform) the default action is dispatched a second time. There's no need for this, and in some forms it leads to an error (identifier already exists).

Note: The forms behave a bit differently because actions are set up differently. When pressing enter, the role and content type edit forms save and reload. The section and content type group forms save and close.

https://jira.ez.no/browse/EZP-24942

Also required: https://github.com/ezsystems/PlatformUIBundle/pull/403


andrerom:
Should consider changing signature of `dispatchFormAction` to mark that actionName can be null (incl default value to null)

glye:
@andrerom Thanks, done.

pspanja:
+1

andrerom:
+1

PR #54, opened 2015-10-26
Fix EZP-25002: Unable to add more options to ezselection field type

glye:
This is the simplest solution, with no javascript. It adds a button "Add option / remove empty options" which will remove any empty options, and add a new one if none are empty. Will reload the form for each press of the button.

Screenshot

Dynamic editing via javascript is a bit more complex, see http://symfony.com/doc/current/cookbook/form/form_collections.html#cookbook-form-collections-new-prototype

https://jira.ez.no/browse/EZP-25002

### DIFF
--- a/bundle/Resources/translations/ezrepoforms_content_type.en.xlf
+++ b/bundle/Resources/translations/ezrepoforms_content_type.en.xlf
@@ -279,6 +279,10 @@
                 <source>field_definition.ezselection.options</source>
                 <target>Options</target>
             </trans-unit>
+            <trans-unit id="9346e6056cfea40f0386d670b68b7450">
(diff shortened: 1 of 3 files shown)


glye:
Review ping @dpobel @mhyndle @andrerom @bdunogier @yannickroger

bdunogier:
Why is the same button used for 2 things (add/remove empty) ?

glye:
@bdunogier Because this is the quickest possible solution. I wanted to run this through review first. If it's not accepted, then the logical next step is a javascript implementation with one add button, and a remove buttion for each option, which will take more time.

bdunogier:
I'm not dead set against it. I'd rather have a functional but midly satisfying solution than no solution at all :-)

pspanja:
+1

PR #55, opened 2015-10-27
Fix EZP-24569: Automatically compute FieldDefinition position

glye:
> Implements https://jira.ez.no/browse/EZP-24569
> Status: Ready for review
> BDD for this: https://github.com/ezsystems/PlatformUIBundle/pull/423

Changes:
- Fielddefs are sorted when the content type is updated
- New fielddefs are created with position = max existing position + 1, so they show up last
- When the position is the same, sort by ~~create order~~ identifier (alphabetically ascending)

Tasks
- [x] "max existing position + 1" doesn't work when a position has been edited at the same time (same submit) as a new fielddef is added - the edited position is not taken into account
- [x] Tests

### DIFF
--- a/lib/Data/ContentTypeData.php
+++ b/lib/Data/ContentTypeData.php
@@ -44,4 +44,22 @@ class ContentTypeData extends ContentTypeUpdateStruct implements NewsnessCheckab
     {
         $this->fieldDefinitionsData[] = $fieldDefinitionData;
     }
+
+    /**
+     * Sort $this->fieldDefinitionsData first by position, then by identifier.
+     */
+    public function sortFieldDefinitions()
+    {
+        usort(
+            $this->fieldDefinitionsData,
+            function ($a, $b) {
+                if ($a->fieldDefinition->position === $b->fieldDefinition->position) {
+                    // The identifiers can never be the same
+                    return $a->fieldDefinition->identifier < $b->fieldDefinition->identifier ? -1 : 1;
+                }
+
+                return $a->fieldDefinition->position < $b->fieldDefinition->position ? -1 : 1;
+            }
+        );
+    }
 }
--- a/lib/Data/Mapper/ContentTypeDraftMapper.php
+++ b/lib/Data/Mapper/ContentTypeDraftMapper.php
@@ -60,6 +60,7 @@ class ContentTypeDraftMapper implements FormDataMapperInterface
                 'isSearchable' => $fieldDef->isSearchable,
             ]));
         }
+        $contentTypeData->sortFieldDefinitions();
 
         return $contentTypeData;
     }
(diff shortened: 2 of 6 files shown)


glye:
Added test for new method `ContentTypeData::sortFieldDefinitions()`.

bdunogier:
I'm wondering if we should try to cover this with BDD. It's PJAX, meaning that we can ignore the whole PlatformUI juggernaut, and it should work pretty neatly.

Thoughts, M. Lye ?

glye:
@bdunogier If you mean *instead of* the currently failing test then I'm all for, struggling a bit with this. But that might be famous last words from me, haven't tried making BDD from scratch. So can't really comment.

andrerom:
+1

lib/Data/ContentTypeData.php
```diff
@@ -44,4 +44,25 @@ public function addFieldDefinitionData(FieldDefinitionData $fieldDefinitionData)
     {
         $this->fieldDefinitionsData[] = $fieldDefinitionData;
     }
+
+    /**
+     * Sort $this->fieldDefinitionsData first by position, then by creation order.
```
pspanja: ID is not really equal to creation order, that is incidentally the case now with Legacy Storage Engine, but with another engine the ID could be a hash string. Hence it would be better to use `identifier`, as this won't change between storage engines (and even between export/import).

glye: @pspanja Yep, agreed. CC @andrerom (just to clarify, `priority` would still be the first sort criterion)

glye:
Changed to use `identifier` as second sort criterium insted of `id`.

pspanja:
+1

andrerom:
+1

glye:
Added a sorting call to `ContentTypeDraftMapper::mapToFormData()`, solving a case where sorting didn't work.

glye:
@lolautruche I think this should have your blessing too, before I merge. It's your issue, and this solution differs from what you wanted:
- "Position needs to be computed from the order FieldDefinitions are created in, unless the position has been forced with a defined value." - See above, we changed this to sort by identifier, since there is no timestamp, and the autoincremented ID is legacy specific. But new fielddefs now have priority set to max() + 1, so they appear last.
- "Ideally positions need to be calculated on the client side." - This solution is server side only.

lib/Data/ContentTypeData.php
```diff
@@ -44,4 +44,25 @@ public function addFieldDefinitionData(FieldDefinitionData $fieldDefinitionData)
...
+    /**
+     * Sort $this->fieldDefinitionsData first by position, then by identifier.
+     */
+    public function sortFieldDefinitions()
+    {
+        usort(
+            $this->fieldDefinitionsData,
+            function ($a, $b) {
+                if ($a->fieldDefinition->position === $b->fieldDefinition->position) {
+                    if ($a->fieldDefinition->identifier === $b->fieldDefinition->identifier) {
```
lolautruche: This cannot happen as we can't have the same field definition identifier within the same ContentType

lolautruche:
Besides my remark, +1

lib/Data/ContentTypeData.php
glye: True, I guess that's paranoia on my end.

lolautruche: So maybe you can remove this condition?

glye: Sure, coming up asap, unless our internet connection drops out again.

glye:
Removed unneeded sorting clause, thanks @lolautruche

PR #56, opened 2015-10-30
Fix EZP-25002: Unable to add more options to ezselection field type

glye:
> Fixes https://jira.ez.no/browse/EZP-25002
> Also requires https://github.com/ezsystems/PlatformUIBundle/pull/417
> Status: Ready for review

Dynamic JS-based solution with separate add/remove buttons.

IScreenshot

### DIFF
--- a/bundle/Resources/views/ContentType/field_types.html.twig
+++ b/bundle/Resources/views/ContentType/field_types.html.twig
@@ -183,18 +183,34 @@
         {{- form_widget(form.isMultiple) -}}
     </div>
 
-    <div class="ezselection-settings options{% if group_class is not empty %} {{ group_class }}{% endif %}">
+    <div id="{{form.options.vars.id}}-ezselection-root-options-list"
+         class="ezselection-settings options{% if group_class is not empty %} {{ group_class }}{% endif %}"
+         data-prototype="{{ ( form_widget(form.options.vars.prototype) ~ ' ' ~
+                              '<button id="' ~ form.options.vars.id ~ '-ez-selection-delete-option-button-__prototype__"' ~
+                              '        class="pure-button ez-font-icon ez-button ez-button-delete ez-selection-delete-option-button"' ~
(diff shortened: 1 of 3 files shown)


glye:
Ready for review, ping @mhyndle @dpobel @yannickroger

glye:
(Added commit from Friday, which syncs with changes in the other PR.)

andrerom:
+1

bundle/Resources/views/ContentType/field_types.html.twig
```diff
@@ -183,18 +183,34 @@
         {{- form_widget(form.isMultiple) -}}
     </div>
 
-    <div class="ezselection-settings options{% if group_class is not empty %} {{ group_class }}{% endif %}">
+    <div id="{{form.options.vars.id}}-ezselection-root-options-list"
+         class="ezselection-settings options{% if group_class is not empty %} {{ group_class }}{% endif %}"
+         data-prototype="{{ ( form_widget(form.options.vars.prototype) ~ ' ' ~
```
dpobel: no! we can not do that, that's kind of ugly. But I think the whole topic deserves a bit more thinking/dev.

glye: @dpobel Do you mean the concept of making prototype html and store it in an html attribute, and use javascript to write it into the document? That's the way Symfony recommends it:
http://symfony.com/doc/current/cookbook/form/form_collections.html#cookbook-form-collections-new-prototype

Or do you mean my implementation of it? That may of course be ugly :) I am concatenating more stuff manually. It might be possible to put all of it in a twig template and get Symfony to accept that as prototype instead of the default.

dpobel: I'm talking about the concept. I need to think a bit more about the topic, but instead of storing the *prototype* into a data attribute, I would rather store it in a `<script>` tag and call that a template. Also, we could maybe replace the ugly `String.replace(__whatever__)` by a proper template usage.

bdunogier:
Looks like the change will happen later, I'll close this for the time being.

PR #57, opened 2015-11-10
Fix EZP-25079: When saving a new policy draft module/function choice is lost

lolautruche:
> https://jira.ez.no/browse/EZP-25079

When saving a policy without publishing, we now retrieve the updated role and inject
the new policy into the data object.

PR #58, opened 2015-11-11
Fix EZP-25021: Implement missing limitation mappers

glye:
> Implements https://jira.ez.no/browse/EZP-25021
> State: Ready for review

Adds the missing limitation mappers, used when editing role policies.

**TODO**
- [x] Owner
- [x] Group
- [x] ~~Node (will need UDW)~~ Moved to https://github.com/ezsystems/repository-forms/pull/63
- [x] ~~Subtree (will need UDW)~~ Moved to https://github.com/ezsystems/repository-forms/pull/63
- [x] State
- [x] NewState
- [x] ParentOwner
- [x] ParentGroup
- [x] ParentClass
- [x] ParentDepth
- [x] Language
- [x] NewSection

### DIFF
--- a/bundle/Resources/config/role.yml
+++ b/bundle/Resources/config/role.yml
@@ -22,6 +22,12 @@ parameters:
     ezrepoforms.limitation.form_mapper.siteaccess.class: EzSystems\RepositoryForms\Limitation\Mapper\SiteAccessLimitationMapper
     ezrepoforms.limitation.form_mapper.contenttype.class: EzSystems\RepositoryForms\Limitation\Mapper\ContentTypeLimitationMapper
     ezrepoforms.limitation.form_mapper.section.class: EzSystems\RepositoryForms\Limitation\Mapper\SectionLimitationMapper
+    ezrepoforms.limitation.form_mapper.objectstate.class: EzSystems\RepositoryForms\Limitation\Mapper\ObjectStateLimitationMapper
+    ezrepoforms.limitation.form_mapper.language.class: EzSystems\RepositoryForms\Limitation\Mapper\LanguageLimitationMapper
+    ezrepoforms.limitation.form_mapper.owner.class: EzSystems\RepositoryForms\Limitation\Mapper\OwnerLimitationMapper
+    ezrepoforms.limitation.form_mapper.group.class: EzSystems\RepositoryForms\Limitation\Mapper\GroupLimitationMapper
+    ezrepoforms.limitation.form_mapper.parentdepth.class: EzSystems\RepositoryForms\Limitation\Mapper\ParentDepthLimitationMapper
+    ezrepoforms.limitation.form_mapper.parentdepth.maxdepth: 20
 
 services:
     # Form types
@@ -95,9 +101,79 @@ services:
         tags:
             - { name: ez.limitation.formMapper, limitationType: Class }
 
+    ezrepoforms.limitation.form_mapper.parentcontenttype:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.contenttype.class%
+        arguments: [@ezpublish.api.service.content_type]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: ParentClass }
+
     ezrepoforms.limitation.form_mapper.section:
         parent: ezrepoforms.limitation.form_mapper.multiple_selection
         class: %ezrepoforms.limitation.form_mapper.section.class%
         arguments: [@ezpublish.api.service.section]
         tags:
             - { name: ez.limitation.formMapper, limitationType: Section }
+
+    ezrepoforms.limitation.form_mapper.newsection:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.section.class%
+        arguments: [@ezpublish.api.service.section]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: NewSection }
+
+    ezrepoforms.limitation.form_mapper.objectstate:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.objectstate.class%
+        arguments: [@ezpublish.api.service.object_state]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: State }
+
+    ezrepoforms.limitation.form_mapper.newobjectstate:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.objectstate.class%
+        arguments: [@ezpublish.api.service.object_state]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: NewState }
+
+    ezrepoforms.limitation.form_mapper.language:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.language.class%
+        arguments: [@ezpublish.api.service.language]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Language }
+
+    ezrepoforms.limitation.form_mapper.owner:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.owner.class%
+        arguments: [@translator]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Owner }
+
+    ezrepoforms.limitation.form_mapper.parentowner:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.owner.class%
+        arguments: [@translator]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: ParentOwner }
+
+    ezrepoforms.limitation.form_mapper.group:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.group.class%
+        arguments: [@translator]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Group }
+
+    ezrepoforms.limitation.form_mapper.parentgroup:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.group.class%
+        arguments: [@translator]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: ParentGroup }
+
+    ezrepoforms.limitation.form_mapper.parentdepth:
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.parentdepth.class%
+        arguments: [%ezrepoforms.limitation.form_mapper.parentdepth.maxdepth%]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: ParentDepth }
--- a/lib/Limitation/Mapper/GroupLimitationMapper.php
+++ b/lib/Limitation/Mapper/GroupLimitationMapper.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+use eZ\Publish\API\Repository\Values\User\Limitation;
+use Symfony\Component\Translation\TranslatorInterface;
+
+class GroupLimitationMapper extends MultipleSelectionBasedMapper
+{
+    /**
+     * @var TranslatorInterface
+     */
+    private $translator;
+
+    public function __construct(TranslatorInterface $translator)
+    {
+        $this->translator = $translator;
+    }
+
+    protected function getSelectionChoices()
+    {
+        return [
+            1 => $this->translator->trans('policy.limitation.group.self', [], 'ezrepoforms_role'),
+        ];
+    }
+}
--- a/lib/Limitation/Mapper/OwnerLimitationMapper.php
+++ b/lib/Limitation/Mapper/OwnerLimitationMapper.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+use eZ\Publish\API\Repository\Values\User\Limitation;
+use Symfony\Component\Translation\TranslatorInterface;
+
+class OwnerLimitationMapper extends MultipleSelectionBasedMapper
+{
+    /**
+     * @var TranslatorInterface
+     */
+    private $translator;
+
+    public function __construct(TranslatorInterface $translator)
+    {
+        $this->translator = $translator;
+    }
+
+    protected function getSelectionChoices()
+    {
+        // 2: "Session" is not supported yet, see OwnerLimitationType
+        return [
+            1 => $this->translator->trans('policy.limitation.owner.self', [], 'ezrepoforms_role'),
+        ];
+    }
+}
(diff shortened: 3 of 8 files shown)


glye:
Not tested yet due to https://github.com/ezsystems/repository-forms/pull/57#discussion_r44527363

glye:
@lolautruche Work in progress, but please see the questions in the descriptions above.

glye:
Added Node mapper which is still work in progress, but functioning now. (I think I'll make a `UDWBasedMapper` base class for Node and Subtree, depending on how different they'll have to be given the different input requirements.)

lolautruche:
> Shouldn't all these mappers be optional? ("required" = false)

I don't think so. You always need to configure the limitations, like it was the case before in legacy. If you don't need to configure the limitation, then you probably don't need a limitation.

bundle/Resources/config/role.yml
```diff
@@ -95,9 +105,71 @@ services:
         tags:
             - { name: ez.limitation.formMapper, limitationType: Class }
 
+    ezrepoforms.limitation.form_mapper.parentcontenttype:
+        parent: ezrepoforms.limitation.form_mapper.contenttype
+        class: %ezrepoforms.limitation.form_mapper.parentcontenttype.class%
+        arguments: [@ezpublish.api.service.content_type]
```
lolautruche: This is not needed as you already defined `ezrepoforms.limitation.form_mapper.contenttype` as parent.

bundle/Resources/config/role.yml
```diff
@@ -95,9 +105,71 @@ services:
...
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.owner.class%
+        arguments: [@translator]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Owner }
+
+    ezrepoforms.limitation.form_mapper.parentowner:
+        parent: ezrepoforms.limitation.form_mapper.owner
+        class: %ezrepoforms.limitation.form_mapper.parentowner.class%
+        arguments: [@translator]
```
lolautruche: Same as above. No need to redefine arguments.

bundle/Resources/config/role.yml
```diff
@@ -95,9 +105,71 @@ services:
...
+        parent: ezrepoforms.limitation.form_mapper.multiple_selection
+        class: %ezrepoforms.limitation.form_mapper.group.class%
+        arguments: [@translator]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Group }
+
+    ezrepoforms.limitation.form_mapper.parentgroup:
+        parent: ezrepoforms.limitation.form_mapper.group
+        class: %ezrepoforms.limitation.form_mapper.parentgroup.class%
+        arguments: [@translator]
```
lolautruche: Same

bundle/Resources/views/Limitation/udw_limitation_value.html.twig
```diff
@@ -0,0 +1,14 @@
...
+{{ form_widget(form.limitationValues) }}
+
+<button data-universaldiscovery-title="{{"role.policy.limitation.location.udw_title"|trans({}, "ezrepoforms_role")}}"
+        class="ez-button-tree pure-button ez-font-icon ez-button ez-pick-location-limitation-button"
+        data-location-input-selector="#{{form.limitationValues.vars.id}}"
+        data-selected-location-selector="#{{form.limitationValues.vars.id}}-selected-location"
+>{{"role.policy.limitation.location.udw_button"|trans({}, "ezrepoforms_role")}}</button>
+<div id="{{form.limitationValues.vars.id}}-selected-location">
+    {% if form.limitationValues.vars.value is not empty %}
+        {{ render( controller( "ez_content:viewLocation", {'locationId': form.limitationValues.vars.value, 'viewType': '_platformui_content_name'} ) ) }}
```
lolautruche: Deprecation warning!

You should now use `ez_content:view`, passing `contentId` instead of `locationId`. `viewType` is still valid.

lib/Limitation/Mapper/GroupLimitationMapper.php
```diff
@@ -0,0 +1,53 @@
...
+    public function __construct(TranslatorInterface $translator)
+    {
+        $this->translator = $translator;
+    }
+
+    protected function getSelectionChoices()
+    {
+        return [
+            0 => $this->translator->trans('policy.limitation.group.any', [], 'ezrepoforms_role'),
+            1 => $this->translator->trans('policy.limitation.group.self', [], 'ezrepoforms_role'),
```
lolautruche: Why enforcing the keys? Not needed AFAICS.

lib/Limitation/Mapper/OwnerLimitationMapper.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->translator = $translator;
+    }
+
+    protected function getSelectionChoices()
+    {
+        return [
+            0 => $this->translator->trans('policy.limitation.owner.any', [], 'ezrepoforms_role'),
+            1 => $this->translator->trans('policy.limitation.owner.self', [], 'ezrepoforms_role'),
+            // 2: "Session" is not supported yet, see OwnerLimitationType
+            //2 => $this->translator->trans('policy.limitation.owner.session', [], 'ezrepoforms_role'),
```
lolautruche: Same as above

lib/Limitation/Mapper/OwnerLimitationMapper.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    protected function getChoiceFieldOptions()
+    {
+        return ['required' => false];
+    }
+
+    public function filterLimitationValues(Limitation $limitation)
+    {
+        // 0: "Any" cannot be selected. We don't store it as a limitation value because it's the same as no limitation
```
lolautruche: Then don't add "any" to the list, like for CT or Section

lib/Limitation/Mapper/GroupLimitationMapper.php
glye: @lolautruche Not strictly needed to enforce the keys, but IMHO being verbose here has value for others reading the code - the keys correspond to specific values in the limitationtype code. Might consider if we should use constants, though.

glye:
@lolautruche 
> I don't think so. You always need to configure the limitations, like it was the case before in legacy. If you
> don't need to configure the limitation, then you probably don't need a limitation.

> Then don't add "any" to the list, like for CT or Section

I wasn't being clear, sorry. Taken together, these suggestions lead to a problem: If all the settings are required, and we don't have an "Any" choice for Owner and Group, then any content / read policy MUST have owner: "self" and group: "self" set. Which is a problem :smile: (In the current platformui, at least. With a different UI it might be worked around.)

lolautruche:
As far as I saw in legacy code, *Any* was always dumped. For ContentType, if you don't select anything, it will act as "any".

glye:
@lolautruche Yes, but that's the thing... if the limitation is not optional, you *can't* not select anything. We can *either* have required parameters, *or* use selects without a "no choice/any" option. Not both :smiley:

lolautruche:
This is why `required: false` is mostly used

lolautruche:
What I meant by saying that limitation config is not optional was for the mappers, not for the user.

glye:
Ok, then it's clear :thumbsup:

glye:
The Node mapper is working now. (The Subtree mapper should be much of the same.)

lib/Limitation/Mapper/LocationLimitationMapper.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public function getFormTemplate()
+    {
+        return $this->template;
+    }
+
+    public function filterLimitationValues(Limitation $limitation)
+    {
+        $limitation->limitationValues = [$limitation->limitationValues];
```
lolautruche: Errr, isn't there something wrong here? :smile:

lolautruche:
I guess you should now implement an abstract UDW mapper

lib/Limitation/Mapper/LocationLimitationMapper.php
glye: Possibly, but it works... The form returns the value directly, but the limitation wants it in an array. So I wrap it here, and unwrap it in `mapLimitationForm()`.

glye:
> abstract UDW mapper

Node and Subtree don't have that much in common now, but ok.

lolautruche:
They at least share the same template

glye:
Added `NewSection` and `NewState` mappers.
Added abstract UDW mapper for the `Node` and `Subtree` mappers.
Found a bug when leaving the UDW mappers unset, working on that.

glye:
Review ping @lolautruche @andrerom @bdunogier et al.

andrerom:
+1

lib/Limitation/Mapper/ContentTypeLimitationMapper.php
```diff
@@ -33,4 +33,9 @@ protected function getSelectionChoices()
 
         return $contentTypeChoices;
     }
+
+    protected function getChoiceFieldOptions()
+    {
+        return ['required' => false];
```
lolautruche: If we have to add `['required' => false]` everywhere, maybe we should add this as default in abstract mapper

lib/Limitation/Mapper/NewObjectStateLimitationMapper.php
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+class NewObjectStateLimitationMapper extends ObjectStateLimitationMapper
```
lolautruche: If you dont have anything to do here, maybe the class is not needed. Just refer to `ObjectStateLimitationMapper` in service definition

lib/Limitation/Mapper/NewSectionLimitationMapper.php
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+class NewSectionLimitationMapper extends SectionLimitationMapper
```
lolautruche: Same as NewObjectStateLimitationMapper

lib/Limitation/Mapper/LocationLimitationMapper.php
```diff
@@ -0,0 +1,19 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+use eZ\Publish\API\Repository\Values\User\Limitation;
+
+class LocationLimitationMapper extends UDWBasedMapper
+{
+    public function filterLimitationValues(Limitation $limitation)
```
lolautruche: I don't see this method used in `USWBasedMapper`. Am I missing something?

lib/Limitation/Mapper/OwnerLimitationMapper.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $this->translator = $translator;
+    }
+
+    protected function getSelectionChoices()
+    {
+        return [
+            1 => $this->translator->trans('policy.limitation.owner.self', [], 'ezrepoforms_role'),
+            // 2: "Session" is not supported yet, see OwnerLimitationType
+            //2 => $this->translator->trans('policy.limitation.owner.session', [], 'ezrepoforms_role'),
```
lolautruche: Please don't commit commented out code. It's fine to have a comment (should be above `return` statement), with what it should be, but commented out code is evil :wink:

lib/Limitation/Mapper/ParentContentTypeLimitationMapper.php
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+class ParentContentTypeLimitationMapper extends ContentTypeLimitationMapper
```
lolautruche: Same as for `NewObjectStateMapper`

lib/Limitation/Mapper/ParentDepthLimitationMapper.php
```diff
@@ -0,0 +1,38 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+class ParentDepthLimitationMapper extends MultipleSelectionBasedMapper
+{
+    /**
+     * @var int The maximum possible depth to use in a limitation
+     * (In legacy the max depth is set in site.ini [RoleSettings] MaxParentDepthLimitation, default: 20)
```
lolautruche: This is not legacy :smiley: . I think you can remove this comment part.

lib/Limitation/Mapper/ParentGroupLimitationMapper.php
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+class ParentGroupLimitationMapper extends GroupLimitationMapper
```
lolautruche: Same as `NewObjectStateMapper`

lib/Limitation/Mapper/ParentOwnerLimitationMapper.php
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Limitation\Mapper;
+
+class ParentOwnerLimitationMapper extends OwnerLimitationMapper
```
lolautruche: Same as `NewObjectStateMapper`

lib/Limitation/Mapper/LocationLimitationMapper.php
lolautruche: Forget it, I'm stupid. I just forgot my own code :blush:

glye: You mean you'd like an empty implementation there, like in `MultipleSelectionBasedMapper`? Sure, can do. Is it CS/policy to implement every part of an interface, even in an abstract class?

lib/Limitation/Mapper/NewObjectStateLimitationMapper.php
glye: Cool, then the same goes for `ParentContentTypeLimitationMapper`, `ParentGroupLimitationMapper`, and `ParentOwnerLimitationMapper`. Checking...

lib/Limitation/Mapper/LocationLimitationMapper.php
glye: Oh, ok :)

lib/Limitation/Mapper/NewObjectStateLimitationMapper.php
lolautruche: Less is more :-)

glye:
Set `['required' => false]` as default in the abstract mapper, and removed it from subclasses
Removed empty classes, and updated service definition accordingly
Removed commented code
Removed legacy comment

lolautruche:
+1

bdunogier:
So, can we merge this ?

glye:
When https://github.com/ezsystems/PlatformUIBundle/pull/433 is approved, yes.

bdunogier:
How much does this one depend on ezsystems/platform-ui-bundle#433 ?

glye:
Moved the UDW-based mappers (Node and Subtree) to new PR, since https://github.com/ezsystems/PlatformUIBundle/pull/433 will take a while.

bdunogier:
And we're in. If not already done, remember to change the issue # in #63, @glye.

PR #59, opened 2015-11-16
EZP-25100: Content edition support in repository-forms

lolautruche:
> https://jira.ez.no/browse/EZP-25100

Adds content edition support.

Basically, content form consists of several sub-forms, one for each content field (as defined by the ContentType). Each FieldType needs a mapper in order to add editable form fields to the content form. For this, `FieldTypeFormMapperInterface` has been updated with a new method:

```php
    /**
     * "Maps" Field form to current FieldType.
     * Allows to add form fields for content edition.
     *
     * @param FormInterface $fieldForm Form for the current Field.
     * @param FieldData $data Underlying data for current Field form.
     */
    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data);
```

For convenience, an `AbstractMapper` has been also added in order to ease the development of such a mapper.
Example for `TextBlockFormMapper`:

```php
namespace EzSystems\RepositoryForms\FieldType\Mapper;

use EzSystems\RepositoryForms\Data\Content\FieldData;
use EzSystems\RepositoryForms\Data\FieldDefinitionData;
use Symfony\Component\Form\FormInterface;

class TextBlockFormMapper extends AbstractMapper
{
    // ...

    /**
     * Returns the form FieldType to be used for content editing.
     *
     * @return string|\Symfony\Component\Form\FormTypeInterface
     */
    protected function getContentFormFieldType()
    {
        return 'textarea';
    }

    /**
     * Returns options to be passed to the form FieldType used in content edition context.
     *
     * @param FormInterface $fieldForm
     * @param FieldData $data
     *
     * @return array
     */
    protected function getContentFormFieldTypeOptions(FormInterface $fieldForm, FieldData $data)
    {
        return [
            'attr' => ['rows' => $data->fieldDefinition->fieldSettings['textRows']],
        ];
    }
}
```

If the FieldType is more complex than just a simple and single form field, best is to [create a custom form field type](http://symfony.com/doc/current/cookbook/form/create_custom_field_type.html) (see `UserType` for instance).

If no mapper is implemented for a given FieldType and it is used in a content, corresponding field will appear as non-editable.

## Implemented mappers
* TextLine
* TextBlock
* User

## TODO
* [x] Implement repository-forms logic for content editing / publishing (action dispatcher + form processor)
* [x] Implement basic `ContentEditController`
* [x] Implement a few mappers
* [ ] Hook in the new view system in order to make the ContentEditController reusable


bundle/Controller/ContentEditController.php
```diff
@@ -0,0 +1,146 @@
...
+
+use eZ\Bundle\EzPublishCoreBundle\Controller;
+use eZ\Publish\API\Repository\ContentService;
+use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\LocationService;
+use EzSystems\RepositoryForms\Data\Mapper\ContentCreateMapper;
+use EzSystems\RepositoryForms\Data\Mapper\ContentUpdateMapper;
+use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
+use EzSystems\RepositoryForms\Form\Type\Content\ContentCreateType;
+use EzSystems\RepositoryForms\Form\Type\Content\ContentEditType;
```
bdunogier: `ContentEditType` / `ContentUpdateMapper` ? `Create` is consistent across both, can we try to get edit closer as well ? I suppose that "Update" follows the repository's naming, but maybe...

lib/FieldType/Mapper/AbstractMapper.php
```diff
@@ -0,0 +1,95 @@
...
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use eZ\Publish\API\Repository\FieldTypeService;
+use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
+use EzSystems\RepositoryForms\Data\Content\FieldData;
+use EzSystems\RepositoryForms\FieldType\DataTransformer\FieldValueTransformer;
+use EzSystems\RepositoryForms\FieldType\FieldTypeFormMapperInterface;
+use Symfony\Component\Form\FormInterface;
+
+abstract class AbstractMapper implements FieldTypeFormMapperInterface
```
bdunogier: Hmm, while reading the mapper, I'm thinking that we might wanna reverse the logic here

Instead of having every Mapper extend this one, why don't we make the common behaviour a service that each mapper uses ? It avoids the tight coupling we get with protected methods.

bdunogier:
Looks globally okay to me, +1.

andrerom:
Besides `ContentEditType` => `ContentUpdateType`, and missing implementations of field types, it looks sane to me as well.

However would be helpful to know what you consider remaining from your POV.

lolautruche:
@andrerom @bdunogier It's called `ContentEditType` because this form type is meant for both creation (without draft) and update.

lib/Form/Type/Content/ContentEditType.php
```diff
@@ -0,0 +1,52 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Form\Type\Content;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class ContentEditType extends AbstractType
```
andrerom: class doc to reflect this is used for both updates and creation w/o draft

andrerom:
> It's called ContentEditType because this form type is meant for both creation (without draft) and update.

ok, then added remark on adding class doc on this.

Anything missing/planned further from your side?

lolautruche:
@andrerom Added the comment on ContentEditType

lolautruche:
@andrerom @bdunogier updated the description

lolautruche:
Added validation and fixed missing default value when creating.

lib/Data/Content/ContentUpdateData.php
```diff
@@ -0,0 +1,37 @@
...
+
+use eZ\Publish\Core\Repository\Values\Content\ContentUpdateStruct;
+use EzSystems\RepositoryForms\Data\NewnessCheckable;
+
+/**
+ * @property-read \EzSystems\RepositoryForms\Data\Content\FieldData[] $fieldsData
+ * @property-read \eZ\Publish\API\Repository\Values\ContentType\ContentType $contentType
+ * @property-read \eZ\Publish\API\Repository\Values\Content\Content $contentDraft
+ */
+class ContentUpdateData extends ContentUpdateStruct implements NewnessCheckable
```
bdunogier: It would be convenient, from the ContentEditView perspective, if both `ContentUpdateData` and `ContentCreateData` could implement a common interface (`ContentData` ?).

While at it, maybe `ContentData` could expose a `getContentType()` method. I could us that, and it would give this sucker something to do instead of just being there eating memory ;)

lib/FieldType/Mapper/UserFormMapper.php
```diff
@@ -0,0 +1,35 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\Form\Type\UserType;
+use Symfony\Component\Form\FormInterface;
+
+class UserFormMapper extends AbstractMapper
```
bdunogier: Not sure we should have this one at all. User isn't editable via the Content API...

We also can't really use it for registration, since the `UserCreateStruct` doesn't use fields for user account data (it has first level properties for username, email, password...).

bdunogier:
Unless you have something else you want to add now, i think this is _good enough_ for the time being.

+1

bdunogier:
Replaced by #67, with a more limited scope.

PR #61, opened 2015-11-24
EZP-24924: Moved page services to own package

bdunogier:
> Part of http://jira.ez.no/browse/EZP-24924

Moved `PageFormMapper` + service declaration to `ezsystems/ezplatform-page-fieldtype`.


andrerom:
+1

PR #62, opened 2015-11-26
EZP-25100: Added content edit view

bdunogier:
> PR against https://github.com/ezsystems/repository-forms/pull/59
> Adds View support for content edit
> Requires https://github.com/ezsystems/ezpublish-kernel/pull/1517

Form data, and the form in general, is processed by the ContentEditForm Filter.

The ViewBuilder will:
- if the forms needs to be (re)displayed, build a ContentEditView object.
  The request then enters the classic view workflow, where the default controller action is empty, and the template is used in the view listener. The controller action/template will be customizable with the matching ViewProvider and MatcherFactory.
- If the form was valid, build a ContentEditSuccessView.
  This view is processed by an actual controller action, that isn't meant to
  be customized. It uses the actionDispatcher, and if the dispatcher returns a response (redirection), use it as the controller's response.

### TODO
- [x] Add the matcher stuff (to customize the view / set custom templates)
- [x] Add semantic config for `content_edit_view`
- [x] See how matchers can work with this (a ContentEditView is not really a ContentView, since it doesn't really contain a Content. How can we make this work ?)
- [ ] Test the workflow with draft

### DIFF
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -12,11 +12,9 @@ use eZ\Bundle\EzPublishCoreBundle\Controller;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LocationService;
-use eZ\Publish\Core\MVC\Symfony\View\View;
-use EzSystems\RepositoryForms\Data\Mapper\ContentCreateMapper;
 use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\Content\ContentCreateType;
-use EzSystems\RepositoryForms\Form\Type\Content\ContentEditType;
+use EzSystems\RepositoryForms\View\ContentEdit\ContentEditSuccessView;
 use EzSystems\RepositoryForms\View\ContentEdit\ContentEditView;
 use Symfony\Component\HttpFoundation\Request;
 
@@ -54,27 +52,11 @@ class ContentEditController extends Controller
         $this->contentService = $contentService;
     }
 
-    public function createWithoutDraftAction($contentTypeId, $language, $parentLocationId, Request $request)
+    public function createWithoutDraftAction(ContentEditView $view)
     {
-        $contentType = $this->contentTypeService->loadContentType($contentTypeId);
-        $data = (new ContentCreateMapper())->mapToFormData($contentType, [
-            'mainLanguageCode' => $language,
-            'parentLocation' => $this->locationService->newLocationCreateStruct($parentLocationId),
-        ]);
-        $form = $this->createForm(new ContentEditType(), $data, ['languageCode' => $language]);
-        $form->handleRequest($request);
+        $view->setTemplateIdentifier('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig');
 
-        if ($form->isValid()) {
-            $this->contentActionDispatcher->dispatchFormAction($form, $data, $form->getClickedButton()->getName());
-            if ($response = $this->contentActionDispatcher->getResponse()) {
-                return $response;
-            }
-        }
-
-        return $this->render('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig', [
-            'form' => $form->createView(),
-            'languageCode' => $language,
-        ]);
+        return $view;
     }
 
     public function createAction($contentTypeId, $language, $parentLocationId)
@@ -120,6 +102,8 @@ class ContentEditController extends Controller
 
     public function editAction(ContentEditView $view)
     {
+        $view->setTemplateIdentifier('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig');
+
         return $view;
     }
 
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -12,11 +12,12 @@ use eZ\Bundle\EzPublishCoreBundle\Controller;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\Core\MVC\Symfony\View\View;
 use EzSystems\RepositoryForms\Data\Mapper\ContentCreateMapper;
-use EzSystems\RepositoryForms\Data\Mapper\ContentUpdateMapper;
 use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\Content\ContentCreateType;
 use EzSystems\RepositoryForms\Form\Type\Content\ContentEditType;
+use EzSystems\RepositoryForms\View\ContentEdit\ContentEditView;
 use Symfony\Component\HttpFoundation\Request;
 
 class ContentEditController extends Controller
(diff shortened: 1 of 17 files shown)


lib/View/ContentEdit/ContentEditViewBuilder.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        if (!isset($parameters['form']) || !$parameters['form'] instanceof FormInterface) {
+            throw new \InvalidArgumentException("Missing or invalid 'form' view parameter");
+        }
+
+        /** @var FormInterface $form */
+        $form = $parameters['form'];
+
+        if ($form->isValid()) {
+            // @lolautruche: is there an else ?
```
lolautruche: No, else is always implicit. If the condition si fulfilled, we are supposed to redirect.

lolautruche: So I guess you're supposed to return a `ContentEditSuccessView` here

bdunogier:
Updated create without draft to use views. Form processing is done in the Form Filter, and the `ContentEditViewBuilder` will either create a `ContentEditView` for rendering if the form needs to be displayed, or a `ContentEditSuccessView` if processing is finished.

lib/View/ContentEdit/ContentEditViewBuilder.php
bdunogier: Yep, that's what it does now.

bdunogier:
Added a prototype of content type identifier based matching.

Requires a patch to the kernel, but overriding the edit form by content type does work.

bundle/Controller/ContentEditController.php
```diff
@@ -117,30 +102,29 @@ public function createDraftAction(Request $request, $contentId, $language, array
...
-        $data = (new ContentUpdateMapper())->mapToFormData($contentDraft, [
-            'languageCode' => $language,
-            'contentType' => $contentType,
-        ]);
-        $form = $this->createForm(new ContentEditType(), $data, [
-            'languageCode' => $language,
-            'drafts_enabled' => true,
-        ]);
-        $form->handleRequest($request);
+        $view->setTemplateIdentifier('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig');
```
lolautruche: Why not checking if we already have one like in `createWithoutDraftAction()`?

bundle/DependencyInjection/Configuration/Parser/ContentEditView.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function addSemanticConfig(NodeBuilder $nodeBuilder)
+    {
+        $nodeBuilder
+            ->arrayNode('content_edit_view')
+                ->info('View selection settings when displaying a content edit view')
+                ->children()
+                ->arrayNode('full')
+                    ->useAttributeAsKey('key')
```
lolautruche: Please use a relevant name instead of *key* (e.g. `view_name`)

bundle/DependencyInjection/Configuration/Parser/ContentEditView.php
```diff
@@ -0,0 +1,65 @@
...
+                                ->example('subdir/my_template.html.twig')
+                            ->end()
+                            ->scalarNode('controller')
+                                ->info(
+                                    <<<EOT
+Use custom controller instead of the default one to display a content edit matching your rules.
+You can use the controller reference notation supported by Symfony.
+EOT
+                                )
+                                ->example('MyBundle:MyControllerClass:editArticleAction')
```
lolautruche: Example is a bit misleading, since with this notation one must not user `Controller` or `Action` suffixes

bundle/DependencyInjection/Configuration/Parser/ContentEditView.php
```diff
@@ -0,0 +1,65 @@
...
+                            ->end()
+                            ->arrayNode('match')
+                                ->info('Condition matchers configuration')
+                                ->useAttributeAsKey('key')
```
lolautruche: Same as above

lib/View/ContentEdit/ContentEditFormFilter.php
```diff
@@ -0,0 +1,114 @@
...
+        $this->contentService = $contentService;
+        $this->contentTypeService = $contentTypeService;
+        $this->formFactory = $formFactory;
+        $this->locationService = $locationService;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            ViewEvents::FILTER_BUILDER_PARAMETERS => [/*'addContentEditForm', */'addContentEditWithoutDraftForm'],
```
lolautruche: Commented code? :)

lib/View/ContentEdit/ContentEditFormFilter.php
```diff
@@ -0,0 +1,114 @@
...
+            'contentType' => $contentType,
+        ]);
+        $form = $this->formFactory->create(new ContentEditType(), $data, [
+            'languageCode' => $request->attributes->get('language'),
+            'drafts_enabled' => true,
+        ]);
+
+        $e->getParameters()->add([
+            'form', $form->handleRequest($request),
+            'contentTypeId' => $contentType->id
```
lolautruche: ContentType is reachable from the data object. I'd have passed form + data (even though data is accessible with `$form->getData()`

lib/View/ContentEdit/ContentEditFormFilter.php
lolautruche:
```php
return [
    ViewEvents::FILTER_BUILDER_PARAMETERS => [['addContentEditForm'], ['addContentEditWithoutDraftForm']],
];
```

Should work to make the 2 methods work.

lib/View/ContentEdit/ContentEditSuccessView.php
```diff
@@ -0,0 +1,41 @@
...
+use eZ\Publish\Core\MVC\Symfony\View\View;
+use Symfony\Component\Form\Form;
+use Symfony\Component\Form\FormInterface;
+
+class ContentEditSuccessView extends BaseView implements View/*, FormView */
+{
+    /**
+     * @var \Symfony\Component\Form\FormInterface
+     */
+    private $form;
```
lolautruche: Here I'd also pass `$data`

lib/View/ContentEdit/ContentEditSuccessView.php
```diff
@@ -0,0 +1,41 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\View\ContentEdit;
+
+use eZ\Publish\Core\MVC\Symfony\View\BaseView;
+use eZ\Publish\Core\MVC\Symfony\View\View;
+use Symfony\Component\Form\Form;
+use Symfony\Component\Form\FormInterface;
+
+class ContentEditSuccessView extends BaseView implements View/*, FormView */
```
lolautruche: Commented code?

lib/View/ContentEdit/ContentEditSuccessView.php
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * The Form object that was successfully processed.
+     * Could have been a FormInterface, but we use getClickedButton() on the view, and it is not part of the interface.
+     *
+     * @param Form $form A form object.
+     *
+     * @return ContentEditView
+     */
+    public function setForm($form)
```
lolautruche: No type hinting?

lib/View/ContentEdit/ContentEditView.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * @var string
+     */
+    private $language;
+
+    /**
+     * @var mixed
+     */
+    private $contentTypeId;
```
lolautruche: If you inject the data object, you won't need `$contentTypeId` and the setter.

lib/View/ContentEdit/ContentEditViewBuilder.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function buildView(array $parameters)
+    {
+        if (!isset($parameters['form']) || !$parameters['form'] instanceof FormInterface) {
+            throw new \InvalidArgumentException("Missing or invalid 'form' view parameter");
+        }
+
+        if (!isset($parameters['contentTypeId'])) {
+            throw new \InvalidArgumentException("Missing or invalid 'contentTypeId' parameter");
+        }
```
lolautruche: OptionsResolver?

lib/View/ContentEdit/FormViewParameterInjector.php
```diff
@@ -0,0 +1,30 @@
...
+class FormViewParameterInjector implements EventSubscriberInterface
+{
+    public static function getSubscribedEvents()
+    {
+        return [ViewEvents::FILTER_VIEW_PARAMETERS => 'injectForm'];
+    }
+
+    public function injectForm(FilterViewParametersEvent $event)
+    {
+        if (!$event->getView() instanceof FormView) {
```
lolautruche: Can you please assign `$event->getView()` to a variable since you use it below?

lolautruche:
+1 besides comments

lib/View/ContentEdit/ContentEditViewBuilder.php
bdunogier: I'll think about it :)

lib/View/ContentEdit/ContentEditSuccessView.php
bdunogier: Forgot, I guess :)

lib/View/ContentEdit/ContentEditFormFilter.php
bdunogier: Actually, I don't need it anymore, unless I'm mistaken. If you look at the ContentEditViewBuilder, it actually doesn't even use it. Probably an artifact. And if I need it, the view can return it now, anyway.

The builder is also meant to build objects it requires based on parameters using repo services & stuff.

lib/View/ContentEdit/ContentEditSuccessView.php
bdunogier: Well, this might be part of the `FormView` interface (I don't wanna add methods to views that aren't part of an interface).

We could go for getForm / getFormData, but why would we provide the data separately (setter) ? We could as well return what's in the form, as a proxy (I'm not 100% sure what the benefit would be though.

But I'm also gonna be in need of an interface for `ContentEditSuccessView::getFormView()`... we have a tiny name conflict with both us and the Form component implementing a `FormView` interface ;-)

lib/View/ContentEdit/ContentEditFormFilter.php
bdunogier: Yep, I'll add it. But how do I test the other edit thing ? :-)

lib/View/ContentEdit/ContentEditSuccessView.php
bdunogier: See reply above about the `FormView` interface.

bdunogier:
Do you think we should have a distinct story for that one, @lolautruche ?

bundle/Controller/ContentEditController.php
andrerom: ditto

andrerom:
+1 with one open question

bdunogier:
Replaced by #67, with a more limited scope.

PR #64, opened 2015-11-27
Fix EZP-25159: Implement missing limitation mappers Node and Subtree

glye:
> Implements https://jira.ez.no/browse/EZP-25159
> State: Ready for review
> Also requires https://github.com/ezsystems/PlatformUIBundle/pull/433

Adds the missing, UDW-based limitation mappers, used when editing role policies. Extracted from https://github.com/ezsystems/repository-forms/pull/58

- [x] Node
- [x] Subtree
- [x] Multiple selection

### DIFF
--- a/bundle/Resources/config/role.yml
+++ b/bundle/Resources/config/role.yml
@@ -28,6 +28,8 @@ parameters:
     ezrepoforms.limitation.form_mapper.group.class: EzSystems\RepositoryForms\Limitation\Mapper\GroupLimitationMapper
     ezrepoforms.limitation.form_mapper.parentdepth.class: EzSystems\RepositoryForms\Limitation\Mapper\ParentDepthLimitationMapper
     ezrepoforms.limitation.form_mapper.parentdepth.maxdepth: 20
+    ezrepoforms.limitation.udw.template: "EzSystemsRepositoryFormsBundle:Limitation:udw_limitation_value.html.twig"
+    ezrepoforms.limitation.form_mapper.subtree.class: EzSystems\RepositoryForms\Limitation\Mapper\SubtreeLimitationMapper
 
 services:
     # Form types
@@ -177,3 +179,20 @@ services:
         arguments: [%ezrepoforms.limitation.form_mapper.parentdepth.maxdepth%]
         tags:
             - { name: ez.limitation.formMapper, limitationType: ParentDepth }
+
+    ezrepoforms.limitation.form_mapper.udw_based:
+        class: EzSystems\RepositoryForms\Limitation\Mapper\UDWBasedMapper
+        calls:
+            - [setFormTemplate, [%ezrepoforms.limitation.udw.template%]]
+
+    ezrepoforms.limitation.form_mapper.location:
+        parent: ezrepoforms.limitation.form_mapper.udw_based
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Node }
+
+    ezrepoforms.limitation.form_mapper.subtree:
+        parent: ezrepoforms.limitation.form_mapper.udw_based
+        class: %ezrepoforms.limitation.form_mapper.subtree.class%
+        arguments: [@ezpublish.api.service.location]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Subtree }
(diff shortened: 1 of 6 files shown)


bundle/Resources/config/role.yml
```diff
@@ -177,3 +180,18 @@ services:
...
+        calls:
+            - [setFormTemplate, [%ezrepoforms.limitation.udw.template%]]
+
+    ezrepoforms.limitation.form_mapper.subtree:
+        class: %ezrepoforms.limitation.form_mapper.subtree.class%
+        arguments: [@ezpublish.api.service.location]
+        tags:
+            - { name: ez.limitation.formMapper, limitationType: Subtree }
+        calls:
+            - [setFormTemplate, [%ezrepoforms.limitation.udw.template%]]
```
glye: @lolautruche says: As already suggested (IIRC), this call and the same one for the location mapper can be abstracted in an abstract service.

glye:
Added an abstract service that calls the UDW template.

glye:
Added an abstract service that calls the UDW template.
CC @lolautruche

lolautruche:
+1

glye:
As mentioned in daily, review ping @bdunogier @andrerom

bdunogier:
+1

This can be merged before the PlatformUI PR, right ?

glye:
Thanks!

> This can be merged before the PlatformUI PR, right ?

Not really. It adds stuff to a template which won't work without the JS in platformui.

lib/Limitation/Mapper/SubtreeLimitationMapper.php
```diff
@@ -0,0 +1,35 @@
...
+
+    public function __construct(LocationService $locationService)
+    {
+        $this->locationService = $locationService;
+    }
+
+    public function filterLimitationValues(Limitation $limitation)
+    {
+        // UDW returns a location ID. If we haven't used UDW, the value is as stored: a path string.
+        if (preg_match('/\A\d+\z/', $limitation->limitationValues) === 1) {
```
andrerom: Seems that limitationValues is set as a single value, as opposed to documented array, seems like a bit wrong use of the object.

lib/Limitation/Mapper/UDWBasedMapper.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function getFormTemplate()
+    {
+        return $this->template;
+    }
+
+    public function mapLimitationForm(FormInterface $form, Limitation $data)
+    {
+        $form->add('limitationValues', 'hidden', [
+            'data' => $data->limitationValues[0],
```
andrerom: Not supporting several values?

glye: Right, this should be multiple value. Fixing...

glye:
Enabled selection of multiple values per limitation.

andrerom:
+1

lib/Limitation/Mapper/UDWBasedMapper.php
```diff
@@ -0,0 +1,56 @@
...
+                    'required' => false,
+                    'label' => $data->getIdentifier(),
+                ])
+                ->addModelTransformer(new UDWBasedValueTransformer())
+                // Deactivate auto-initialize as we're not on the root form.
+                ->setAutoInitialize(false)->getForm()
+        );
+    }
+
+    public function filterLimitationValues(Limitation $limitation)
```
bdunogier: abstract method + class ?

glye: @bdunogier That's what I used to have, but I removed it because after other changes that class contained only this method and it was empty. So I thought, better to have an empty method here, than an empty class.

bdunogier: Well, better, I don't know. There is another method in there now, is there not ?

glye: @bdunogier The `SubtreeLimitationMapper` uses it, but `LocationLimitationMapper` (the empty class I removed) did not. `LimitationFormMapperInterface` specifies it.

glye:
Renamed location selector, ref platformui PR.

PR #65, opened 2015-12-07
Fix EZP-25203: Pressing <Enter> when creating a role saves the role but don't publish it

glye:
> Fixes https://jira.ez.no/browse/EZP-25203
> Also required: https://github.com/ezsystems/PlatformUIBundle/pull/465
> Status: Ready for review

Let the default action call the save role action. Let the save role action both save and publish.

### DIFF
--- a/lib/Form/Processor/RoleFormProcessor.php
+++ b/lib/Form/Processor/RoleFormProcessor.php
@@ -37,28 +37,61 @@ class RoleFormProcessor implements EventSubscriberInterface
 
     public function processDefaultAction(FormActionEvent $event)
     {
-        // Don't update anything if we just want to cancel the draft.
-        if ($event->getClickedButton() === 'removeDraft') {
-            return;
+        if ($this->isDefaultEvent($event)) {
+            $this->processSaveRole($event);
         }
-
-        /** @var \EzSystems\RepositoryForms\Data\Role\RoleData $roleData */
-        $roleData = $event->getData();
-        $roleDraft = $roleData->roleDraft;
-        $this->roleService->updateRoleDraft($roleDraft, $roleData);
     }
 
     public function processSaveRole(FormActionEvent $event)
     {
-        /** @var RoleDraft $roleDraft */
-        $roleDraft = $event->getData()->roleDraft;
+        $roleDraft = $this->getRoleDraft($event);
+
+        $this->roleService->updateRoleDraft($roleDraft, $this->getRoleData($event));
         $this->roleService->publishRoleDraft($roleDraft);
     }
 
     public function processRemoveDraft(FormActionEvent $event)
     {
-        /** @var RoleDraft $roleDraft */
-        $roleDraft = $event->getData()->roleDraft;
-        $this->roleService->deleteRoleDraft($roleDraft);
+        $this->roleService->deleteRoleDraft($this->getRoleDraft($event));
+    }
+
+    /**
+     * Returns true if the event is the default event, meaning Enter has been pressed in the form.
+     *
+     * There's no need to process the default action (save) for explicit events (save, cancel).
+     * Saving is not needed when cancelling, and when the Save button is clicked the save action takes care of it.
+     * The default action is only needed when the form has been submitted by pressing Enter.
+     *
+     * @param FormActionEvent $event
+     *
+     * @return bool
+     */
+    protected function isDefaultEvent(FormActionEvent $event)
+    {
+        return $event->getClickedButton() === null;
+    }
+
+    /**
+     * Returns the role data for the event.
+     *
+     * @param FormActionEvent $event
+     *
+     * @return \EzSystems\RepositoryForms\Data\Role\RoleData
+     */
+    protected function getRoleData(FormActionEvent $event)
+    {
+        return $event->getData();
+    }
+
+    /**
+     * Returns the role draft for the event.
+     *
+     * @param FormActionEvent $event
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\RoleDraft
+     */
+    protected function getRoleDraft(FormActionEvent $event)
+    {
+        return $this->getRoleData($event)->roleDraft;
     }
 }
--- a/tests/RepositoryForms/Form/Processor/RoleFormProcessorTest.php
+++ b/tests/RepositoryForms/Form/Processor/RoleFormProcessorTest.php
@@ -1,0 +1,103 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Tests\Form\Processor;
+
+use eZ\Publish\Core\Repository\Values\User\RoleDraft;
+use EzSystems\RepositoryForms\Data\Role\RoleData;
+use EzSystems\RepositoryForms\Event\FormActionEvent;
+use EzSystems\RepositoryForms\Event\RepositoryFormEvents;
+use EzSystems\RepositoryForms\Form\Processor\RoleFormProcessor;
+use PHPUnit_Framework_TestCase;
+
+class RoleFormProcessorTest extends PHPUnit_Framework_TestCase
+{
+    /**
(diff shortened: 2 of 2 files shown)


andrerom:
One question; Is `processDefaultAction` always called before the actions that corresponds to a button? I don't see any return or similar logic that says "I'm done, don't execute further events". 
So In other words I don't fully understand this part:

```php
+        // Don't do the default action when saving is specified, as it's the same action and shouldn't be done twice.
+        if (in_array($event->getClickedButton(), ['removeDraft', 'saveRole'])) {
             return;
         }
```

glye:
@andrerom Yes, afaict `processDefaultAction` is always called before the actions that corresponds to a button. So when `saveRole` is pressed, we get here first. Without this return, the role would be published twice, which gives us an error since at the second pass the draft doesn't exist anymore.

When enter is pressed, only the default action runs. The if doesn't match, and the save role action gets run only because I call it from the default action.

crevillo:
Cause disabling the possibility to submit the form pressing return is not an option, right?

andrerom:
> Cause disabling the possibility to submit the form pressing return is not an option, right?

that does not make sense from html / usability perspective.


However, I'm having dubts on this whole thing.
Are we now publishing on enter in most forms? ping @dpobel 
*Cause it is annoying to publish content on a mistaken enter, when you didn't mean to as well*

glye:
Pressing enter in:
Role: Publish
Section: Publish
Content type: Save, not publish
Language: Error (I'll make an issue)

glye:
The state of things in my previous comment makes sense to me: Content type editing is a complex form with potentially harmfull consequences if you make a mistake. OTOH, the Role, Section and Content type group forms are trivial and Mostly Harmless (RIP Douglas Adams) because you're only editing a name.

crevillo:
And what do we want to do on enter key pressed? Can't tabindex be added to buttons so when enter is pressed it "submits" the one with the minimun tabindex?

andrerom:
@crevillo this is true, @glye can you check that? would be cleaner imho then approach here.

glye:
@andrerom @crevillo I tried these `tabIndex` cases, while leaving the `processDefaultAction` empty:
1: text input, 2: save, 3: cancel
1: save, 2: cancel, 3: text input
In both cases, pressing enter means that the form reloads. Neither save nor cancel is called. So to do what you suggest, it seems we may need to involve the JS gang :grinning: 
CC @dpobel @yannickroger @mhyndle

andrerom:
Is the primary button of type `<input type="submit">` and the first one in the dom?

glye:
@andrerom Both Save and Cancel are of type `submit`. Cancel is first, but reversing that doesn't change the result.

andrerom:
ok, well do like this then.
+1

lib/Form/Processor/RoleFormProcessor.php
```diff
@@ -38,20 +38,22 @@ public static function getSubscribedEvents()
     public function processDefaultAction(FormActionEvent $event)
     {
         // Don't update anything if we just want to cancel the draft.
-        if ($event->getClickedButton() === 'removeDraft') {
+        // Don't do the default action when saving is specified, as it's the same action and shouldn't be done twice.
+        if (in_array($event->getClickedButton(), ['removeDraft', 'saveRole'])) {
             return;
         }
 
-        /** @var \EzSystems\RepositoryForms\Data\Role\RoleData $roleData */
```
bdunogier: Naming nitpick, but I'd try to replace this typehinting, done 3 times in the class, with a better named method with proper phpdoc (`$this->getRoleData()` ?).

What would make it the cleanest would be a decoration of the event we get, with methods that know about the underlying data (`$event->getRoleDraft()`, for instance). But not now :-)

lib/Form/Processor/RoleFormProcessor.php
```diff
@@ -38,20 +38,22 @@ public static function getSubscribedEvents()
     public function processDefaultAction(FormActionEvent $event)
     {
         // Don't update anything if we just want to cancel the draft.
-        if ($event->getClickedButton() === 'removeDraft') {
+        // Don't do the default action when saving is specified, as it's the same action and shouldn't be done twice.
+        if (in_array($event->getClickedButton(), ['removeDraft', 'saveRole'])) {
```
bdunogier: I'd abstract it to a boolean test, maybe `$this->skipSave($event)`. It would read a bit better, and would avoid the inline comment (that would then be on the method's phpdoc).

bdunogier:
As far as I could read on the web, if you have only one "submit" button, as opposed to "button" or "cancel", then it will be the one used for the enter key: http://stackoverflow.com/a/16462846/680263 (confirmed by other reads).

Not sure if you have or not tried that.

About the default action, I agree that publishing on enter is too much. But this might even apply to all forms. It's easy to press enter by mistake...

glye:
@bdunogier I tried this (only one submit) but in that case the form just reloads, the submit action isn't called. I guess Symfony affects this somehow. Anyway, I feel safer being specific here rather than trust the browser to do what we want.

If publish on enter is too much, then there is no bug and we can close this PR. As it is, it saves on enter. But all the other admin forms, except content type edit, do publish on enter. All of these are simple one or two line forms with hardly any potential for disaster if you didn't mean to publish. For role, we are only talking about the role name, policies are handled in a different form.

bdunogier:
Okay. As long as it is more consistent, we are headed the right direction.

lib/Form/Processor/RoleFormProcessor.php
glye: Ok. There are 2 different variables, so 2 doc-ed methods added.

lib/Form/Processor/RoleFormProcessor.php
glye: Ok, done.

glye:
Added helper methods to reduce typehinting and inline comment.
Review ping @bdunogier @yannickroger

lib/Form/Processor/RoleFormProcessor.php
```diff
@@ -37,28 +37,62 @@ public static function getSubscribedEvents()
...
+     * Returns true if saving should not be done in the default process.
+     *
+     * There's no need to save when cancelling the draft, nor when the Save button is clicked, as the save process
+     * takes care of it. Saving is only needed when the form has been submitted by pressing Enter.
+     *
+     * @param FormActionEvent $event
+     *
+     * @return bool
+     */
+    protected static function skipDefaultSave(FormActionEvent $event)
```
bdunogier: Why `static` ? As far as I can tell, the methods that use the protected methods are _not_ static.

I'd also call that particular one a bit differently. "skipXXX" sounds like the method will actually _skip_ something. What about "mustSave" ?

glye: @bdunogier Using `static` because I can. Is the best practice to only do it when it actually avoids instantiation?

"skip" was your proposal, man :) But I can change it.

glye: @bdunogier Undid `static`. Renamed helper to `isDefaultEvent`, which made the default action smaller and clearer.

bdunogier:
> @bdunogier Using static because I can. Is the best practice to only do it when it actually avoids instantiation?

> "skip" was your proposal, man :) But I can change it.

Please don't throw their own words at people's face, it's terrible diplomacy. I plead darwin.

bdunogier:
Any unit testing around this ? :-)

That aside, +1.

glye:
Applied the same simplification as @dpobel suggested in the platformui PR, for the default event.

@bdunogier If you put a gun to my head, I'll see what I can do :)

bdunogier:
:headphones: :gun:

That's the best i could do.

bdunogier:
Honestly, writing a unit test for the RoleProcessor doesn't sound very complicated.

glye:
@bdunogier Then I'm glad to hear you can help me out with the RoleFormProcessorTest I just added ;)
It fails at `testProcessDefaultAction`, where `isDefaultEvent` is never called. The other tests pass.

bdunogier:
Never called ? That's pretty odd, is it not ?

glye:
I was using the service mock, but these methods are in the form processor. Pushed new version which also mocks the form processor, still same error.

bdunogier:
Yeah, I was about to say that.

You shouldn't mock the form processor, since you're testing it :-)

glye:
Updated, tests pass locally now.

PR #66, opened 2016-01-08
Fix EZP-25432 In Textline fieldtype, is being permitted the "Minimum length" and "Maximum length" to have negative values

crevillo:
> Part of https://jira.ez.no/browse/EZP-25342
> Related PR https://github.com/ezsystems/ezpublish-kernel/pull/1553

Added a min attribute to the ```input type="number"``` so capable browsers can warn the user if negative value (or ```0```) is entered.

### DIFF
--- a/lib/FieldType/Mapper/TextLineFormMapper.php
+++ b/lib/FieldType/Mapper/TextLineFormMapper.php
@@ -24,11 +24,13 @@ class TextLineFormMapper implements FieldTypeFormMapperInterface
                 'required' => false,
                 'property_path' => 'validatorConfiguration[StringLengthValidator][minStringLength]',
                 'label' => 'field_definition.ezstring.min_length',
+                'attr' => ['min' => 0],
             ])
             ->add('maxLength', 'integer', [
                 'required' => false,
                 'property_path' => 'validatorConfiguration[StringLengthValidator][maxStringLength]',
                 'label' => 'field_definition.ezstring.max_length',
+                'attr' => ['min' => 0],
             ])
             ->add(
                 // Creating from FormBuilder as we need to add a DataTransformer.


crevillo:
ping @dpobel @yannickroger @bdunogier

andrerom:
+1

glye:
The `attr` field type option is not documented, are you sure it works as intended?
http://symfony.com/doc/2.7/reference/forms/types/integer.html

(For browsers that don't support such hinting there should be a validator in place, and as I recall there is. Did you test that, incidentally?)

crevillo:
Manually tested and seems to work. But i see your point. Any way, if you go down in that page you will get this... 
http://symfony.com/doc/current/reference/forms/types/integer.html#read-only
```
The read_only option has been deprecated and will be removed in 3.0. Instead, use the attr option by setting it to an array with a readonly key....
```
So it looks legit to use attr, doesn't it? 

The validation is done in the kernel related pull request. It won't allow negative values.

glye:
That's the 3.0 doc. It's not mentioned in the 2.7 doc. I know the symfony docs are not always complete, but still... @andrerom I'm fine with it if you are.

andrerom:
It is covered for 2.8:
http://symfony.com/doc/2.8/reference/forms/types/integer.html

And we already use it in twig templates in Platform UI, so I'm ok :)

glye:
ok, +1 then!

lib/FieldType/Mapper/TextLineFormMapper.php
```diff
@@ -24,11 +24,13 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
                 'required' => false,
                 'property_path' => 'validatorConfiguration[StringLengthValidator][minStringLength]',
                 'label' => 'field_definition.ezstring.min_length',
+                'attr' => array('min' => 0),
```
bdunogier: Short array notation ?

bdunogier:
+1 the array nitpick aside.

Let's be forward compatible :-)

crevillo:
changed to short array notation, @bdunogier

PR #67, opened 2016-01-12
Primitive user generated content framework

bdunogier:
> Implements [EZP-25100](https://jira.ez.no/browse/EZP-25100)
> Re-created from https://github.com/ezsystems/repository-forms/pull/59 & https://github.com/ezsystems/repository-forms/pull/62
> Status: ready for review
> Requires https://github.com/ezsystems/ezplatform/pull/85
> TL;DR: Implements a controller action that creates content, without using an intermediary draft, using Symfony forms. Does not include validation.

Provides a `ContentEditController` with a `createWithoutDraft` action.

http://ezplatform.loc/content/create/nodraft/1/eng-GB/2  will generate a content create form for a Folder (`1`). Submitting the form will create a Folder in `eng-GB`, below location `2`.

### Scope
- Only the minimal/most simplistic set of FieldTypes are supported: TextLine, and TextBlock.
- No validation is implemented.
- The only action for now is "create and publish", without using a draft (`createWithoutDraftAction`). We believe this use-case is both the simplest to implement and the most common, as it doesn't require that a draft is created into the repository to display a content form.

#### `ez_content_edit` controller + routes
New routes are added within the `/content` namespace, like `/content/create/nodraft/{contentTypeId}/{language}/{parentLocationId}`, that defaults to simple controller actions.

The controller uses an `ActionDispatcher` to process the actual submitted data.

#### The `FieldValueMapperInterface`
The `FieldTypeFormMapperInterface`, responsible for mapping FieldDefinitions, is deprecated, and replaced by `FieldDefinitionFormMapperInterface`. This interface has an evil twin, `FieldValueFormMapperInterface`, that maps Field Values. All interfaces, including the default one, inherit from `FieldFormMapperInterface`.

The `FieldTypeFormMapperRegistry` is also deprecated. A `FieldTypeFormMapperDispatcher` replaces it. It handles both FieldDefinition and FieldValue through a unique `map()` method:

```php
$this->fieldMapper->map($contentItemForm, $fieldData);
$this->fieldMapper->map($contentTypeForm, $fieldDefinitionData);
```

This mapper will do the mapping if there is a mapper, and skip silently if there isn't.

### TODO
- [x] Change `contentTypeId` to `contentTypeIdentifier`
- [x] Rebase/merge master (deps)
- [x] Remove commit "[tmp] Set ezsystems/ezplatform branch" before merging
- [x] Re-consider the changes to FieldMapper interfaces, and see if a cleaner approach, especially regarding the registry, and how it is used.
- [x] Travis behat integration
- [x] Re-credit form related commits
- [x] Behat coverage
- [x] Consider separating the view api integration from the forms part. Use a basic controller with all the pieces in place ?

### DIFF
--- a/bin/.travis/prepare_behat.sh
+++ b/bin/.travis/prepare_behat.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+# File for setting up system for behat testing, just like done in DemoBundle's .travis.yml
+
+# Change local git repo to be a full one as we will reuse it for composer install below
+git fetch --unshallow && git checkout -b tmp_travis_branch
+export BRANCH_BUILD_DIR=$TRAVIS_BUILD_DIR
+export TRAVIS_BUILD_DIR="$HOME/build/ezplatform"
+cd "$HOME/build"
+
+# Checkout meta repo, change the branch and/or remote to use a different ezpublish branch/distro
+git clone --depth 1 --single-branch --branch master https://github.com/ezsystems/ezplatform.git
+cd ezplatform
+
+# Install everything needed for behat testing, using our local branch of this repo
+./bin/.travis/setup_from_external_repo.sh $BRANCH_BUILD_DIR "ezsystems/repository-forms:dev-tmp_travis_branch as 1.0"
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryFormsBundle\Controller;
+
+use eZ\Bundle\EzPublishCoreBundle\Controller;
+use eZ\Publish\API\Repository\ContentService;
+use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\LocationService;
+use EzSystems\RepositoryForms\Data\Mapper\ContentCreateMapper;
+use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
+use EzSystems\RepositoryForms\Form\Type\Content\ContentEditType;
+use Symfony\Component\HttpFoundation\Request;
+
+class ContentEditController extends Controller
+{
+    /**
+     * @var ContentTypeService
+     */
+    private $contentTypeService;
+
+    /**
+     * @var ContentService
+     */
+    private $contentService;
+
+    /**
+     * @var LocationService
+     */
+    private $locationService;
+
+    /**
+     * @var ActionDispatcherInterface
+     */
+    private $contentActionDispatcher;
+
+    public function __construct(
+        ContentTypeService $contentTypeService,
+        ContentService $contentService,
+        LocationService $locationService,
+        ActionDispatcherInterface $contentActionDispatcher
+    ) {
+        $this->contentTypeService = $contentTypeService;
+        $this->locationService = $locationService;
+        $this->contentActionDispatcher = $contentActionDispatcher;
+        $this->contentService = $contentService;
+    }
+
+    /**
+     * Displays and processes a content creation form. Showing the form does not create a draft in the repository.
+     *
+     * @param int $contentTypeIdentifier ContentType id to create
+     * @param string $language Language code to create the content in (eng-GB, ger-DE, ...))
+     * @param int $parentLocationId Location the content should be a child of
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function createWithoutDraftAction($contentTypeIdentifier, $language, $parentLocationId, Request $request)
+    {
+        $contentType = $this->contentTypeService->loadContentTypeByIdentifier($contentTypeIdentifier);
+        $data = (new ContentCreateMapper())->mapToFormData($contentType, [
+            'mainLanguageCode' => $language,
+            'parentLocation' => $this->locationService->newLocationCreateStruct($parentLocationId),
+        ]);
+        $form = $this->createForm(new ContentEditType(), $data, ['languageCode' => $language]);
+        $form->handleRequest($request);
+
+        if ($form->isValid()) {
+            $this->contentActionDispatcher->dispatchFormAction($form, $data, $form->getClickedButton()->getName());
+            if ($response = $this->contentActionDispatcher->getResponse()) {
+                return $response;
+            }
+        }
+
+        return $this->render('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig', [
+            'form' => $form->createView(),
+            'languageCode' => $language,
+        ]);
+    }
+}
--- a/bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
+++ b/bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
@@ -1,0 +1,43 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryFormsBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use LogicException;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Compiler pass to register FieldType form mappers in the mapper dispatcher.
+ */
+class FieldTypeFormMapperDispatcherPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('ezrepoforms.field_type_form_mapper.dispatcher')) {
+            return;
+        }
+
+        $dispatcherDefinition = $container->findDefinition('ezrepoforms.field_type_form_mapper.dispatcher');
+
+        foreach ($container->findTaggedServiceIds('ez.fieldType.formMapper') as $id => $attributes) {
+            foreach ($attributes as $attribute) {
+                if (!isset($attribute['fieldType'])) {
+                    throw new LogicException(
+                        'ez.fieldType.formMapper service tag needs a "fieldType" attribute to identify which field type the mapper is for. None given.'
+                    );
+                }
+
+                $dispatcherDefinition->addMethodCall('addMapper', [new Reference($id), $attribute['fieldType']]);
+            }
+        }
+    }
+}
--- a/bundle/Resources/config/routing.yml
+++ b/bundle/Resources/config/routing.yml
@@ -1,0 +1,4 @@
+ez_content_create_no_draft:
+    path: /content/create/nodraft/{contentTypeIdentifier}/{language}/{parentLocationId}
+    defaults:
+        _controller: ez_content_edit:createWithoutDraftAction
--- a/bundle/Resources/views/Content/content_form.html.twig
+++ b/bundle/Resources/views/Content/content_form.html.twig
@@ -1,0 +1,23 @@
+{% macro display_form(form) %}
+    {{ form_start(form) }}
+
+    {% for fieldForm in form.fieldsData %}
+        <div class="{{ fieldForm.vars.data.fieldDefinition.fieldTypeIdentifier }}-field field-edit">
+            <fieldset>
+                <legend>{{- form_label(fieldForm) -}}</legend>
+                {{- form_errors(fieldForm) -}}
+                {%- if fieldForm.value is defined -%}
+                    {{ form_errors(fieldForm.value) }}
+                    {{ form_widget(fieldForm.value, {"contentData": form.vars.data}) }}
+                {%- else -%}
+                    <p class="non-editable">
+                        {{ "content.field.non_editable"|trans({}, "ezrepoforms_content") }}
+                    </p>
+                    {%- do fieldForm.setRendered() -%}
+                {% endif %}
+            </fieldset>
+        </div>
+    {% endfor %}
+
+    {{ form_end(form) }}
+{% endmacro %}
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.0.x-dev"
+            "dev-master": "1.1.x-dev"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
         "symfony/validator": "~2.7"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.7"
+        "phpunit/phpunit": "~4.7",
+        "matthiasnoback/symfony-dependency-injection-test": "0.*"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -15,14 +15,16 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.7",
-        "matthiasnoback/symfony-dependency-injection-test": "0.*"
+        "matthiasnoback/symfony-dependency-injection-test": "^0.7",
+        "behat/behat": "^3.0"
     },
     "autoload": {
         "psr-4": {
             "EzSystems\\RepositoryFormsBundle\\": "bundle",
             "EzSystems\\RepositoryForms\\": "lib",
             "EzSystems\\RepositoryForms\\Tests\\": "tests/RepositoryForms",
-            "EzSystems\\RepositoryFormsBundle\\Tests\\": "tests/RepositoryFormsBundle"
+            "EzSystems\\RepositoryFormsBundle\\Tests\\": "tests/RepositoryFormsBundle",
+            "EzSystems\\RepositoryForms\\Features\\": "features"
         }
     },
     "extra": {
--- a/features/Context/ContentEdit.php
+++ b/features/Context/ContentEdit.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\Features\Context;
+
+use Behat\Behat\Context\Context;
+use Behat\Behat\Context\SnippetAcceptingContext;
+use Behat\MinkExtension\Context\MinkContext;
+use PHPUnit_Framework_Assert as Assertion;
+
+class ContentEdit extends MinkContext implements Context, SnippetAcceptingContext
+{
+    /**
+     * Name of the content that was created using the edit form. Used to validate that the content was created.
+     * @var string
+     */
+    private $createdContentName;
+
+    /**
+     * @Then /^I should see a folder content edit form$/
+     * @Then /^I should see a content edit form$/
+     */
+    public function iShouldSeeAContentEditForm()
+    {
+        $this->assertSession()->elementExists('css', 'form[name=ezrepoforms_content_edit]');
+    }
+
+    /**
+     * @Then /^I am on the View of the Content that was published$/
+     */
+    public function iAmOnTheViewOfTheContentThatWasPublished()
+    {
+        if (!isset($this->createdContentName)) {
+            throw new \Exception('No created content name set');
+        }
+
+        $page = $this->getSession()->getPage();
+        Assertion::assertTrue($page->has('css', 'span.ezstring-field'));
+        Assertion::assertEquals($this->createdContentName, $page->find('css', 'span.ezstring-field')->getText());
+    }
+
+    /**
+     * @When /^I fill in the folder edit form$/
+     */
+    public function iFillInTheFolderEditForm()
+    {
+        // will only work for single value fields
+        $this->createdContentName = 'Behat content edit @' . microtime(true);
+        $this->fillField('ezrepoforms_content_edit_fieldsData_name_value', $this->createdContentName);
+    }
+
+    /**
+     * @Given /^that I have permission to create folders$/
+     */
+    public function thatIHavePermissionToCreateFolders()
+    {
+        $this->visit('/login');
+        $this->fillField('Username', 'admin');
+        $this->fillField('Password', 'publish');
+        $this->pressButton('Login');
+    }
+}
--- a/features/Context/ContentType.php
+++ b/features/Context/ContentType.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\Features\Context;
+
+use Behat\Behat\Context\Context;
+use Behat\Behat\Context\SnippetAcceptingContext;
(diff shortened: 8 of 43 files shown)


bundle/Controller/ContentEditController.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->contentTypeService = $contentTypeService;
+        $this->locationService = $locationService;
+        $this->contentActionDispatcher = $contentActionDispatcher;
+        $this->contentService = $contentService;
+    }
+
+    public function createWithoutDraftAction(ContentEditView $view)
+    {
+        if ($view->getTemplateIdentifier() === null) {
+            $view->setTemplateIdentifier('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig');
```
crevillo: couldn't template be personalizable as well? i mean, like view templates...

bdunogier: This isn't meant to remain, I'll add that to the description.

If you dig a bit, _it is personalizable`. This particular template is the default one, and it will be removed from there ultimately. The whole thing is processed by the view API, and therefore connected to the override system. With this config, every view should already use this template by default, and it should be possible to change it already.

I just haven't reached the point where you have semantic config for that, but you can get a piece of the idea here: https://github.com/ezsystems/repository-forms/pull/67/files#diff-1b4b073bac6c1d6bfbce74d7a9a7a115R2.

I think I'll keep the override part for a further PR, given that this part already does something. One of the next steps is to allow custom templates/controllers for `content_edit` by means of semantic config.

crevillo: ok. looks good :)

lib/FieldType/FieldTypeFormMapperInterface.php
```diff
@@ -8,27 +8,16 @@
...
-use EzSystems\RepositoryForms\Data\FieldDefinitionData;
-use Symfony\Component\Form\FormInterface;
+namespace EzSystems\RepositoryForms\FieldType;
 
 /**
  * Interface for FieldType form mappers.
  *
  * It maps a FieldType's specifics to editing Forms (e.g. FieldDefinition forms).
+ *
+ * @deprecated since 1.1, will be removed in 2.0. Use FieldTypeFormMapperInterface instead.
```
glye: FieldDefinitionFormMapperInterface

bdunogier:
PR simplified, moved the View API integration out. Now much closer from #59, but without the user part, other fieldtypes, etc.

lib/FieldType/Mapper/TextLineFormMapper.php
```diff
@@ -44,4 +58,32 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
+        $formConfig = $fieldForm->getConfig();
+        $label = $fieldDefinition->getName($formConfig->getOption('languageCode')) ?: reset($fieldDefinition->getNames());
+
+        $fieldForm
+            ->add(
+                $formConfig->getFormFactory()->createBuilder()
+                    ->create(
+                        'value',
+                        'text',
+                        ['required' => $fieldDefinition->isRequired, 'label' => $label]
```
dpobel: while at it, I guess you can add the min length and max length.

lib/FieldType/Mapper/UserFormMapper.php
```diff
@@ -0,0 +1,35 @@
...
+     * - validator configuration,
+     * - field settings
+     * - default value.
+     *
+     * @param FormInterface $fieldDefinitionForm Form for current FieldDefinition.
+     * @param FieldDefinitionData $data Underlying data for current FieldDefinition form.
+     */
+    public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
+    {
+    }
```
dpobel: since it's empty, I guess you can remove this class for now

lib/FieldType/Mapper/TextLineFormMapper.php
bdunogier: How is that done by the way ? :)

dpobel: Unsure actually. There's a `maxlength` attribute on text input  but no `minlength` attribute so for min length bit more complicated. I guess you can skip my comment for now :)

crevillo: @dpobel it seems there is also minlength https://www.w3.org/TR/html5/forms.html#the-maxlength-and-minlength-attributes

bdunogier: According to stackoverflow, support is a bit erratic. I'll see what I can do. If it works in some places, why not, as long as it doesn't break the others :-)

crevillo: True that. There i also saw some people recomending use ```pattern``` instead...

bdunogier:
Test should be passing now, BDD Included.

bin/.travis/install_composer_github_key.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# This is unfortunately only way to avoid Github API limits with Travis on an open source project.
+# Please don't use our token, neither token nor it's user gives access to anything other then authentication, so better if you create your own.
+# PS: The simple obfuscation here is only to avoid Github from detecting this on commits.
+export EZ_GITHUB_TOKEN_A=`echo "574503f4468c5de5820d" | rev`
+export EZ_GITHUB_TOKEN_B=`echo "90cf1c134798de2dae27" | rev`
```
andrerom: not this file please.

see: https://github.com/ezsystems/ezplatform/commit/bf80b0baa87e3bcdf7f0aa967e6ed2ce8ae8ffc1

bdunogier:
Reverted the changes to `FieldTypeFormMapperInterface`. Since those are meant to be implemented by third-party FieldTypes for integration, I considered the method addition a BC break.

It incurs some added complexity, but here it goes: there are now two types of mappers, `FieldDefinitionFormMapper` and `FieldValueFormMapper`, both inheriting from `FieldFormMapperInterface`. A `FieldFormMapperDispatcher` replaces the `FieldTypeFormMapperRegistry`. It has a unique `map()` method that accepts both `FieldData` and `FieldDefinitionData`, and executes the mapping, if there is one.

The ContentType part needs to be adapted, since it still uses the registry, but it doesn't have to be done right now.

Ready for merge as far as I am concerned.

bdunogier:
Still missing any kind of +1. 

Overall, it "just" applies the form framework we had for ContentType and other value objects to Content. The changes to the registry are a bit invasive, but not really complex.

bin/.travis/install_composer_github_key.sh
bdunogier: Do I need to create a token of some sorts, or can I just use the same ? I won't mind if you do the change yourself :-)

andrerom: you can use the same, it's as mentioned meant for use by ez repos :)

bin/.travis/prepare_behat.sh
```diff
@@ -0,0 +1,16 @@
...
+# File for setting up system for behat testing, just like done in DemoBundle's .travis.yml
+
+# Change local git repo to be a full one as we will reuse it for composer install below
+git fetch --unshallow && git checkout -b tmp_travis_branch
+export BRANCH_BUILD_DIR=$TRAVIS_BUILD_DIR
+export TRAVIS_BUILD_DIR="$HOME/build/ezplatform"
+cd "$HOME/build"
+
+# Checkout meta repo, change the branch and/or remote to use a different ezpublish branch/distro
+git clone --depth 1 --single-branch --branch ugc_base https://github.com/ezsystems/ezplatform.git
```
yannickroger: You should add a TODO in the PR to remove the branch name.

bundle/Controller/ContentEditController.php
```diff
@@ -0,0 +1,76 @@
...
+        LocationService $locationService,
+        ActionDispatcherInterface $contentActionDispatcher
+    ) {
+        $this->contentTypeService = $contentTypeService;
+        $this->locationService = $locationService;
+        $this->contentActionDispatcher = $contentActionDispatcher;
+        $this->contentService = $contentService;
+    }
+
+    public function createWithoutDraftAction($contentTypeId, $language, $parentLocationId, Request $request)
```
yannickroger: Missing phpdoc

bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+namespace EzSystems\RepositoryFormsBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use LogicException;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Compiler pass to register FieldType form mappers in the mapper registry.
```
yannickroger: registry or dispatcher ?

bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+class FieldTypeFormMapperDispatcherPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('ezrepoforms.field_type_form_mapper.dispatcher')) {
+            return;
+        }
+
+        $registry = $container->findDefinition('ezrepoforms.field_type_form_mapper.dispatcher');
```
yannickroger: should this be called dispatcher ?

bundle/Resources/views/Content/content_form.html.twig
```diff
@@ -0,0 +1,23 @@
+{% macro display_form(form) %}
```
yannickroger: Could we put this file in a `macros` folder in order to reflect it only contains a macro and not an actual displayable form ?

lib/FieldType/DataTransformer/FieldValueTransformer.php
```diff
@@ -0,0 +1,48 @@
...
+     * @var FieldType
+     */
+    private $fieldType;
+
+    public function __construct(FieldType $fieldType)
+    {
+        $this->fieldType = $fieldType;
+    }
+
+    public function transform($value)
```
yannickroger: phpdoc here and method bellow

lib/FieldType/FieldDefinitionFormMapperInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+namespace EzSystems\RepositoryForms\FieldType;
+
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use Symfony\Component\Form\FormInterface;
+
+/**
+ * Interface for FieldType form mappers.
+ *
+ * It maps a FieldType's specifics to editing Forms (e.g. FieldDefinition forms).
```
yannickroger: Maybe you could reword this block a bit, it's a bit confusing to me.

lib/FieldType/FieldFormMapperInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
```
yannickroger: Different header from all other files ?

lib/FieldType/FieldTypeFormMapperDispatcher.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+namespace EzSystems\RepositoryForms\FieldType;
+
+use EzSystems\RepositoryForms\Data\Content\FieldData;
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use Symfony\Component\Form\FormInterface;
+use InvalidArgumentException;
+
+/**
+ * Registry for FieldType form mappers.
```
yannickroger: Dispatcher ?

lib/FieldType/FieldTypeFormMapperDispatcherInterface.php
```diff
@@ -0,0 +1,35 @@
...
+use EzSystems\RepositoryForms\Data\Content\FieldData;
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use Symfony\Component\Form\FormInterface;
+
+/**
+ * Registry for FieldType form mappers.
+ */
+interface FieldTypeFormMapperDispatcherInterface
+{
+    public function addMapper(FieldFormMapperInterface $mapper, $fieldTypeIdentifier);
```
yannickroger: missing phpdoc

lib/FieldType/FieldTypeFormMapperDispatcherInterface.php
```diff
@@ -0,0 +1,35 @@
...
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\FieldType;
+
+use EzSystems\RepositoryForms\Data\Content\FieldData;
+use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use Symfony\Component\Form\FormInterface;
+
+/**
+ * Registry for FieldType form mappers.
```
yannickroger: Dispatcher ?

lib/FieldType/FieldTypeFormMapperDispatcher.php
```diff
@@ -0,0 +1,85 @@
...
+                E_USER_DEPRECATED
+            );
+        } elseif (!$mapper instanceof FieldValueFormMapperInterface && !$mapper instanceof FieldDefinitionFormMapperInterface) {
+            throw new \InvalidArgumentException('Expecting a FieldValueFormMapperInterface or FieldDefinitionFormMapperInterface');
+        }
+
+        $this->mappers[$fieldTypeIdentifier] = $mapper;
+    }
+
+    public function addMappers(array $mappersArray)
```
yannickroger: Why are the MapperS methods not part of the `Interface`

bdunogier: Added that yesterday, it's not even used yet. Having this method allows us to add one big method call in the compiler pass, with an array of N mappers, instead of N method calls of 1 mapper. 

I think i'll add it to the interface. Even though somehow, I'd like this better if the `addMapper(s)` methods were in another interface than `map`.

lib/FieldType/FieldValueFormMapperInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
```
yannickroger: Not the same header file as everywhere.

lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -0,0 +1,125 @@
...
+use eZ\Publish\API\Repository\Values\Content\ContentStruct;
+use eZ\Publish\Core\MVC\Symfony\Routing\UrlAliasRouter;
+use EzSystems\RepositoryForms\Event\FormActionEvent;
+use EzSystems\RepositoryForms\Event\RepositoryFormEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+class ContentFormProcessor implements EventSubscriberInterface
```
yannickroger: phpdoc

bundle/Resources/views/Content/content_form.html.twig
bdunogier: I don't think it really matters... this template will ultimately be overridable, meaning that it should have a clear path. But I'll keep the option in mind.

lib/Form/Type/Content/ContentCreateType.php
```diff
@@ -0,0 +1,36 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\Form\Type\Content;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class ContentCreateType extends AbstractType
```
yannickroger: Is this used anywhere ? Looks like ContentEditType is doing the create.

tests/RepositoryForms/FieldType/FieldTypeFormMapperDispatcherTest.php
```diff
@@ -0,0 +1,85 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
```
yannickroger: uncomplete header

tests/RepositoryForms/FieldType/FieldTypeFormMapperDispatcherTest.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * @var \EzSystems\RepositoryForms\FieldType\FieldValueFormMapperInterface|\PHPUnit_Framework_MockObject_MockObject
+     */
+    private $fieldValueMapperMock;
+
+    public function setUp()
+    {
+        $this->dispatcher = new FieldTypeFormMapperDispatcher();
+
+        // Should be improved to test with a value + definition mapper (this is what repo-forms uses)
```
yannickroger: Follow up story ? TODO in the PR ?

tests/RepositoryFormsBundle/DependencyInjection/Compiler/FieldTypeFormMapperPassTest.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
```
yannickroger: Header

yannickroger:
+1 with comments taken into account

lib/Form/Processor/ContentFormProcessor.php
bdunogier: I've added a phpdoc for the class itself. The methods do exactly what they say, and the parameters (the event) are self-explanatory, imho. When I have more to add, I will :-)

tests/RepositoryForms/FieldType/FieldTypeFormMapperDispatcherTest.php
bdunogier: Well, I'd rather go for a `@todo` in the code, but the symfony guidelines don't like those...

yannickroger: Those tend to be forgotten until we have a bug coming up :)

composer.json
```diff
@@ -14,14 +14,17 @@
...
+        "behat/behat": "^3.0"
     },
     "autoload": {
         "psr-4": {
             "EzSystems\\RepositoryFormsBundle\\": "bundle",
             "EzSystems\\RepositoryForms\\": "lib",
             "EzSystems\\RepositoryForms\\Tests\\": "tests/RepositoryForms",
-            "EzSystems\\RepositoryFormsBundle\\Tests\\": "tests/RepositoryFormsBundle"
+            "EzSystems\\RepositoryFormsBundle\\Tests\\": "tests/RepositoryFormsBundle",
+            "EzSystems\\RepositoryForms\\Features\\": "features"
```
andrerom: several of these can be in autoload-dev right?

bundle/Resources/config/routing.yml
```diff
@@ -0,0 +1,4 @@
+ez_content_create_no_draft:
+    path: /content/create/nodraft/{contentTypeId}/{language}/{parentLocationId}
```
andrerom: would probably be more user friendly to use content type identifier?

lib/Data/Mapper/ContentUpdateMapper.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * Maps a ValueObject from eZ content repository to a data usable as underlying form data (e.g. create/update struct).
+     *
+     * @param ValueObject|\eZ\Publish\API\Repository\Values\Content\Content $contentDraft
+     * @param array $params
+     *
+     * @return ContentUpdateData
+     */
+    public function mapToFormData(ValueObject $contentDraft, array $params = [])
```
andrerom: Side as it affects all mappers, and hence should be dealt with separated: *Given the generic nature of the type hint on these, they should probably check type inline before using the properties on the assumed objects.*

andrerom:
+1

I assume this is designed to be able to generate the forms w/o any user data in them so they can be cached somewhat, hence why we rely on kernel to check permission when form is sent, and not when form is created. However one thing we can add some dev doc on is how developers should handle csrf and captcha.

features/Context/ContentEdit.php
```diff
@@ -0,0 +1,63 @@
...
+class ContentEdit extends MinkContext implements Context, SnippetAcceptingContext
+{
+    /**
+     * Name of the content that was created using the edit form. Used to validate that the content was created.
+     * @var string
+     */
+    private $createdContentName;
+
+    /**
+     * @Then /^I should see a folder content edit form$/
```
miguelcleverti: Why use this notation? Why not just `@Then I should see a folder content edit form` ?

features/Context/ContentType.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @Given /^there is a Content Type "([^"]*)" with the id "([^"]*)"$/
+     */
+    public function thereIsAContentTypeWithId($contentTypeIdentifier, $id)
+    {
+        try {
+            $contentType = $this->contentTypeService->loadContentTypeByIdentifier($contentTypeIdentifier);
+            Assertion::assertEquals($id, $contentType->id);
+        } catch (NotFoundException $e) {
+            Assertion::fail("No ContentType with the identifier '$contentTypeIdentifier'' could be found.");
```
crevillo: nitpick ```'$contentTypeIdentifier''```, is the last quote needed?

composer.json
bdunogier: I didn't even know this existed :-)

bundle/Resources/config/routing.yml
bdunogier: > And "create" might kind of imply "nodraft" right?

Hmm does it ? In the PAPI at least it implies creating a draft, if I am not mistaken. 
Semantically speaking, it would be createAndPublish, as it does both in one operation.

bdunogier:
> I assume this is designed to be able to generate the forms w/o any user data in them so they can be cached somewhat

Yes, exactly. The use-case we went for here is a simple content creation form, without drafts or edition. We could of course check for permissions in the controller. It does raise a few potentially interesting questions...

features/Context/ContentEdit.php
bdunogier: I love regular expressions, they're so sexy :-)

bdunogier:
About the route, I'm going to change `contentTypeId` to `contentTypeIdentifier`.

I won't change the route right now, but I'll keep this in mind when working on the others. We aren't releasing this yet, and in any case it is easy to stay backward compatible on routes.

bdunogier:
Should be ready for merge. I'll wait for Travis to pass.

bdunogier:
Added https://jira.ez.no/browse/EZP-25435 for permission check.

andrerom:
Cool!

About the permission issue, reason why I did not suggest we do that is that it then won't be cached easily because of ParentOwner content create limitation, which is when used per user.

But we should add stories for defining best practice for dealing with csrf and captcha.

emodric:
`ez_content_create_no_draft` route is available to anonymous users.

While they can't create the content because of missing `content/create` permissions, shouldn't the form itself be also behind the firewall?

andrerom:
@emodric thanks for the input, I have created a story to track this, check it and provide further input there https://jira.ez.no/browse/EZP-25447

emodric:
@andrerom +1 :)

PR #71, opened 2016-02-08
Fix EZP-25339: Default value length should be between min and max string length

crevillo:
Fixes https://jira.ez.no/browse/EZP-25339

This a requirement to https://github.com/ezsystems/ezpublish-kernel/pull/1574

ping @bdunogier @andrerom

### DIFF
--- a/lib/FieldType/Mapper/TextLineFormMapper.php
+++ b/lib/FieldType/Mapper/TextLineFormMapper.php
@@ -51,6 +51,7 @@ class TextLineFormMapper implements FieldTypeFormMapperInterface, FieldValueForm
                 $fieldDefinitionForm->getConfig()->getFormFactory()->createBuilder()
                     ->create('defaultValue', 'text', [
                         'required' => false,
+                        'property_path' => 'validatorConfiguration[StringLengthValidator][defaultValue]',
                         'label' => 'field_definition.ezstring.default_value',
                     ])
                     ->addModelTransformer(new TextLineValueTransformer())


yannickroger:
you can also ping @glye :)

glye:
+1

bdunogier:
+1

andrerom:
@crevillo ?

crevillo:
wops, didn't see @bdunogier's comment. good question. i also thought about this. i see your point, but if you don't check that property against the StringLengthValidator, how do you check it? I mean, yes, it's a setting, but the strlen of that setting should be between min and max length, shouldn't it?

andrerom:
>  I mean, yes, it's a setting, but the strlen of that setting should be between min and max length, shouldn't it?

It should, but this is not what `property_path ` does afaik: http://symfony.com/doc/current/reference/forms/types/form.html#property-path

property path is afaik wrong here, as the name maps perfectly  to the property it represents: `->defaultValue`.

Checking it against min and max seem to have to be done using something else.

glye:
Agree that defaultValue isn't validator configuration, but a field setting. As for how to check it then, it seems that `FieldDefinitionDefaultValueValidator` exists for this purpose...? (Unsure how to implement it though.)

crevillo:
yep, it's a field setting, but we won't allow the length of this field setting be out of min - max field settings, right? 

Question is... how can i get the posted defaultValue inside https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/TextLine/Type.php#L48? 

Actually, what i noticed about this issue (without this patch) is the following. If you set
 * minLength 2
 * maxLength 6
 * defaultValue 'eZ Platform'

validation passed ok. but after, if you add another field, no matter which type is, then it says that defaultValue can't be greater than the maxLength...

here is the screencast. https://youtu.be/VNG54qmL3yQ

So, why that validation doesn't act at first but seems to act after?

glye:
Correct me if I'm wrong, but I think you're not supposed to validate the defaultValue inside validateValidatorConfiguration(), because it isn't part of the validator. It should be done in validateFieldSettings(). So then the question is, how to read the min/max values from there? :)

andrerom:
Guys, afaik this is not a field setting, it is default value, which is a native property on FieldDefintion.

*I'm assuming we are talking about the default value for the field, as set on field definition, not the default value of max/min for field defintion, as set on $validatorConfigurationSchema.*

And afaik there is nothing around to validate it, if it was a field setting it would most likely have been `validateFieldSettings`.

crevillo:
@gyle agree. the reason of trying to solve it in validateValidatorConfiguration is because i can read those min/max values there :). But i was not really convinced that this was the best way to go.

so, any other ideas?

crevillo:
closing as this is not the way to go.

glye:
Sorry, I don't know. If it isn't a field setting, and the api doesn't support default value validation, then it requires some more thought. (But FieldDefinitionDefaultValueValidator sounds like it was intended for this purpose, at least.)

andrerom:
> (But FieldDefinitionDefaultValueValidator sounds like it was intended for this purpose, at least.)

So that is `EzSystems\RepositoryForms\Validator\Constraints\FieldDefinitionDefaultValueValidator`, indeed, looks relevant.

andrerom:
But it uses `$fieldType->validateValue($fieldDefinition, $fieldValue)` which I can't find anywhere.

emodric:
It is in `eZ\Publish\API\Repository\FieldType`

andrerom:
> It is in eZ\Publish\API\Repository\FieldType

right you are, I was grepping on older checkout :sweat_smile: 

ok, that is the way then @crevillo, it basically inits a field value with the default value on it underneath, so everything gets validated.

crevillo:
That's the reason why that validation does not act at first, then?

crevillo:
i'm still stumble on this. afaik, when you add the textLine type, as you guys said, ```FieldDefinitionDefaultValueValidator``` enters the game. And finally it reach Type Validate function https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/TextLine/Type.php#L80

But looks like at this very first time, ```$constraints``` values are still not set, and hence, the validation returns correct. So, it looks like this validate function that indeed checks the entered defaultValue is called before $constraints minString and maxLenght are set... 

so, any ideas on how to follow from this? @andrerom, @glye, @pspanja

glye:
> But looks like at this very first time, $constraints values are still not set

If you're sure about this, then it seems bigger changes are needed. Hard to suggest anything without digging deeply into it. The first thing I would look into is changing the order of FieldDefinitionDefaultValueValidator and min/max constraints setup, obviously. Any ideas, @bdunogier?

crevillo:
Well, not entirely sure but seems so if you look at the screencast i added https://youtu.be/VNG54qmL3yQ
doing some debug, looks when validate is performed at first constraints is an array with empty minStringLenght and empty maxStringLength.  That's why validate doesn't return errors there. But if you add any other field, then those constraints values are filled in that validate function and hence, the check is performed...

PR #72, opened 2016-02-10
EZP-25376: implemented FieldsGroups in ContentType edition

bdunogier:
> Implements http://jira.ez.no/browse/EZP-25376
> Integrates fields groups added in https://github.com/ezsystems/ezpublish-kernel/pull/1575
> Requires ezpublish-kernel:6.3@dev (FieldsGroups API)

Integrates fields groups in ContentType edit forms. Backward compatibility is covered by injecting `fields_groups.list` service using a setter (avoids adding another optional argument to the ctor), and making it optional in the code.

### Tasks
- [x] Merge https://github.com/ezsystems/ezpublish-kernel/pull/1575 first, and remove the deps commit

### DIFF
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -63,6 +63,8 @@ services:
     ezrepoforms.field_definition.form_type:
         class: %ezrepoforms.field_definition.form_type.class%
         arguments: [@ezrepoforms.field_type_form_mapper.registry, @ezpublish.api.service.field_type]
+        calls:
+            - [setGroupsList, [@ezpublish.fields_groups.list]]
         tags:
             - { name: form.type, alias: ezrepoforms_fielddefinition_update }
 
@@ -224,6 +226,8 @@ services:
     ezrepoforms.form_processor.content_type:
         class: %ezrepoforms.form_processor.content_type.class%
         arguments: [@ezpublish.api.service.content_type, @router, %ezrepoforms.form_processor.content_type.options%]
+        calls:
+            - [setGroupsList, [@ezpublish.fields_groups.list]]
         tags:
             - { name: kernel.event_subscriber }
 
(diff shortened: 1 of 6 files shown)


bdunogier:
[Added](https://github.com/ezsystems/repository-forms/commit/59329743447cba8e306756bc6dc6510cad366c7f) human-readable / multilingual field categories using the translation component.

Might be better to add support for this in https://github.com/ezsystems/ezpublish-kernel/pull/1575. `getGroups()` could be changed to return a hash with `value => label` (the form component expects that anyway), using the translation component internally. It would make sure that labels are centralized, and not dependent on the service's consumer.

andrerom:
_(+1, but @glye and @yannickroger are going to review this with more familiarity)_

glye:
Do we need a separate translation file for field group? I'd expect to see the contenttype file used for this.

bdunogier:
> Do we need a separate translation file for field group?

Probably not, good point.

yannickroger:
+1
I tested it and it seems to work fine.

glye:
> They're not only translations, they're translations of user data.

Ah, "user data" as in "defined in config files". Ok. What if you need to translate a field group related string in a content type template - would that go into the content type file or the field group one? Maybe the file needs an xml comment about what it's for?

(Agree on the prefix if moved to content_type.)

bdunogier:
> What if you need to translate a field group related string in a content type template - would that go into the content type file or the field group one?

content type. field_group is _only_ use to get a human readable version of fields groups identifiers. 

Good as it is then ?

andrerom:
Regarding translation it would fit in ezrepoforms_content_type.en.xlf right now as this is the domain it will be used, but when js UI is translatable it should also be reused(?) for content editing ui, so maybe @dpobel has some input on this.

glye:
> Good as it is then?

If this usage is clear for everyone, yes :) If not, doc it somehow.

bdunogier:
I'll make that clear on the doc.

PR #77, opened 2016-05-04
Fix EZP-25750: textblock content edit error

bdunogier:
> Fixes [EZP-25750](https://jira.ez.no/browse/EZP-25750)

A wrong parameter was passed to the FieldType service thing. Also changed the form widget type that is used to textarea instead of text.

Not much to say, it just fixes it :-)


andrerom:
+1

PR #81, opened 2016-05-12
Fix EZP-25775: Checkbox content edition support

glye:
> Implements https://jira.ez.no/browse/EZP-25775
> Status: Ready for review

Pretty much the simplest possible mapper.

### DIFF
--- a/lib/FieldType/Mapper/CheckboxFormMapper.php
+++ b/lib/FieldType/Mapper/CheckboxFormMapper.php
@@ -1,0 +1,58 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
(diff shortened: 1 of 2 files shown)


lib/FieldType/Mapper/CheckboxFormMapper.php
```diff
@@ -0,0 +1,58 @@
...
+     * @var \eZ\Publish\API\Repository\FieldTypeService
+     */
+    private $fieldTypeService;
+
+    public function __construct(FieldTypeService $fieldTypeService)
+    {
+        $this->fieldTypeService = $fieldTypeService;
+    }
+
+    /**
```
bdunogier: this phpdoc ain't required as it is a copy of the parent's.

glye: Right, removed.

glye:
Moving to feature branch.

PR #82, opened 2016-05-13
EZP-25775: Checkbox content edition support

glye:
> Implements https://jira.ez.no/browse/EZP-25775
> Status: Ready for review

Pretty much the simplest possible mapper.

Moved to feature branch from master: https://github.com/ezsystems/repository-forms/pull/81


lib/FieldType/Mapper/CheckboxFormMapper.php
```diff
@@ -0,0 +1,51 @@
...
+     * @var \eZ\Publish\API\Repository\FieldTypeService
+     */
+    private $fieldTypeService;
+
+    public function __construct(FieldTypeService $fieldTypeService)
+    {
+        $this->fieldTypeService = $fieldTypeService;
+    }
+
+    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
```
jjCavalleri: The signature of mapFieldValueForm method does not match the used signature in other mappers (for instance: [IntegerFormMapper](https://github.com/ezsystems/repository-forms/blob/master/lib/FieldType/Mapper/IntegerFormMapper.php#L19), [TextLineFormMapper](https://github.com/ezsystems/repository-forms/blob/master/lib/FieldType/Mapper/TextLineFormMapper.php#L34))
Would you agree on changing the variable name from **$fieldForm** to **$fieldDefinitionForm**?

bdunogier:
Detail: no need for "fix", it is not an issue :-)

That aside, I have tested the branch, and I'm getting an array to string conversion error if the selection isn't a multiple one. It works if it

bdunogier:
Note: we can squash this when merging, imho.

yannickroger:
+1 once passes also

PR #83, opened 2016-05-13
User FieldType content edit support (EZP-27530)

bdunogier:
> Implements [EZP-25730](http://jira.ez.no/browse/EZP-25730)
> Status: ready for review
> Topic from feature #85

### What it does
- adds a User custom FormType
- adds a generic `FormTypeBasedFieldValueFormMapper`. Can be used for any FieldValueMapper that doesn't require specific configuration, and only uses a FormType, native or custom.
- deprecates the service tag `ez.fieldType.formMapper`, and introduces `ez.fieldFormMapper.value` + `ez.fieldFormMapper.definition`
- adds a Value FieldFormMapper for User, using the generic class
- ~~adds [BDD testing](https://github.com/ezsystems/repository-forms/blob/ezp25730-user_account_content_edit_support/features/FieldEdit/edit_user_field.feature) of the user implementation~~ (moved to the feature branch)
- ~~also adds BDD testing for the textline fieldtype, based on scenarii similar to those of user above~~ (moved to the feature branch)

### What it does not do
Save user data. This requires that we use the UserService instead of the ContentService (like we do in PlatformUI / REST).

### TODO
- [ ] Create follow-up to actually be able to save a user (User vs Content API)
- [x] Consider extracting the textline fieldtype edit form feature + relevant contexts to a new PR. It can be used for existing types, as well as for newly developed ones. Furthermore, the PR is growing a bit large.


glye:
Issue does not exist, says Jira. Guess you can close this, then ;)

bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
```diff
@@ -28,16 +28,44 @@ public function process(ContainerBuilder $container)
...
+                $dispatcherDefinition->addMethodCall('addMapper', [new Reference($id), $tag['fieldType']]);
             }
         }
     }
+
+    /**
+     * Gathers services tagged as either ez.fieldFormMapper.value or ez.fieldFormMapper.definition, as well
+     * as the deprecated ez.fieldType.formMapper.
+     * 
+     * A
```
yannickroger: yes ? :smile:

yannickroger:
+1

features/FieldTypeForm/user_fieldtype_edit_form.feature
```diff
@@ -0,0 +1,22 @@
...
+    As an integrator
+    I want the ezuser field form to implement the FieldType's behaviour
+
+Background:
+    Given a Content Type with a user field definition
+
+Scenario: The attributes of a user field have a form representation
+    When I view the edit form for this field
+    Then the edit form should contain a fieldset named after the field definition
+     And it should contain the following set of labels and input fields types:
```
glye: Micro-nitpick: "input field types"

lib/FieldType/FieldTypeFormMapperDispatcher.php
```diff
@@ -25,10 +25,20 @@ class FieldTypeFormMapperDispatcher implements FieldTypeFormMapperDispatcherInte
     /**
      * FieldType form mappers, indexed by FieldType identifier.
      *
-     * @var FieldTypeFormMapperInterface[]
+     * @var FieldTypeFormMapperInterface[]|FieldValueFormMapperInterface[]|FieldDefinitionFormMapperInterface[]
```
glye: Why this change, when in addMapper() below it looks like only `FieldTypeFormMapperInterface` will ever be used here?

lib/FieldType/Mapper/FormTypeBasedFieldValueFormMapper.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->fieldTypeService = $fieldTypeService;
+    }
+
+    public function setFormType($formType)
+    {
+        $this->formType = $formType;
+    }
+
+    /**
+     * "Maps" Field form to current FieldType.
```
glye: Why the quotation marks? (If `map` isn't really the right word here, then it shouldn't be used in the method name either.) I see the interfaces also quote it, looks wrong there too :)

glye:
+1 apart from (possibly senseless) remarks.

lib/FieldType/Mapper/FormTypeBasedFieldValueFormMapper.php
bdunogier: Good question... I'm not sure why the quotation mark. Maybe I was drowning in the sarcasm sea when I wrote that :)

lib/FieldType/FieldTypeFormMapperDispatcher.php
bdunogier: Because `FieldTypeFormMapperInterface` is deprecated, split into `FieldValue` and `FieldDefinition`. Without it, the value part isn't known, and PhpStorm / IDEs report errors on usage of the value related methods.

features/FieldTypeForm/user_fieldtype_edit_form.feature
bdunogier: Actually, I am not sure. There are several fields, each with their own type...

lib/FieldType/Mapper/FormTypeBasedFieldValueFormMapper.php
glye: Accidental...

features/FieldTypeForm/user_fieldtype_edit_form.feature
glye: Rethinking... the form contains fields, not field types. So better as just "set of labels and input fields".
Or: "the following set of labels, and input fields of the following types" for maximum precision :)

lib/FieldType/FieldTypeFormMapperDispatcher.php
glye: Ok, thanks.

bdunogier:
Fixed remarks, will now rebase & merge.

features/FieldTypeForm/user_fieldtype_edit_form.feature
bdunogier: They're not "fields types", they're "input fields" types. Big difference. We have a domain clash, here, between our content model and the HTML forms domain.

PR #84, opened 2016-05-13
EZP-25776: Selection content edition support

glye:
> Implements https://jira.ez.no/browse/EZP-25776
> Status: Ready for review

Uses a transformer since single/multi selects return integers and arrays, respectively.
- [x] Test the transformer
- [x] BDD

### DIFF
--- a/features/Context/FieldTypeFormContext.php
+++ b/features/Context/FieldTypeFormContext.php
@@ -8,18 +8,19 @@ use Behat\Behat\Context\SnippetAcceptingContext;
 use Behat\Behat\Hook\Scope\BeforeScenarioScope;
 use Behat\Gherkin\Node\TableNode;
 use Behat\Mink\Element\NodeElement;
-use Behat\MinkExtension\Context\MinkContext;
+use Behat\MinkExtension\Context\RawMinkContext;
 use eZ\Publish\API\Repository\Values\ContentType\FieldDefinitionCreateStruct;
 use eZ\Publish\API\Repository\Values\ContentType\FieldDefinitionUpdateStruct;
 use PHPUnit_Framework_Assert as Assertion;
 
-final class FieldTypeFormContext extends MinkContext implements SnippetAcceptingContext
+final class FieldTypeFormContext extends RawMinkContext implements SnippetAcceptingContext
 {
     private static $fieldIdentifier = 'field';
 
     private static $fieldTypeIdentifierMap = [
         'user' => 'ezuser',
         'textline' => 'ezstring',
+        'selection' => 'ezselection',
     ];
 
     /** @var \EzSystems\RepositoryForms\Features\Context\ContentType */
@@ -59,7 +60,7 @@ final class FieldTypeFormContext extends MinkContext implements SnippetAccepting
      */
     public function iEditOrCreateContentOfThisType()
     {
-        $this->visit(
+        $this->visitPath(
             sprintf(
                 '/content/create/nodraft/%s/eng-GB/2',
                 $this->contentTypeContext->getCurrentContentType()->identifier
@@ -72,7 +73,8 @@ final class FieldTypeFormContext extends MinkContext implements SnippetAccepting
      */
     public function theEditFormShouldContainAFieldsetNamedAfterTheFieldDefinition()
     {
-        $this->assertElementContainsText(
+        $this->assertSession()->elementTextContains(
+            'css',
             sprintf('div.ezfield-identifier-%s fieldset legend', self::$fieldIdentifier),
             'Field'
         );
@@ -83,7 +85,8 @@ final class FieldTypeFormContext extends MinkContext implements SnippetAccepting
      */
     public function itShouldContainAGivenTypeInputField($inputType)
     {
-        $this->assertElementOnPage(
+        $this->assertSession()->elementExists(
+            'css',
             sprintf(
                 'div.ezfield-identifier-%s fieldset input[type=%s]',
                 self::$fieldIdentifier,
@@ -110,8 +113,12 @@ final class FieldTypeFormContext extends MinkContext implements SnippetAccepting
                 if ($fieldExpectation['type'] === $nodeElement->getAttribute('type')) {
                     $inputId = $nodeElement->getAttribute('id');
 
-                    $this->assertElementOnPage("label[for=$inputId]");
-                    $this->assertElementContainsText("label[for=$inputId]", $fieldExpectation['label']);
+                    $this->assertSession()->elementExists('css', "label[for=$inputId]");
+                    $this->assertSession()->elementTextContains(
+                        'css',
+                        "label[for=$inputId]",
+                        $fieldExpectation['label']
+                    );
 
                     unset($fieldsExpectations[$expectationId]);
                     reset($fieldsExpectations);
@@ -148,9 +155,7 @@ final class FieldTypeFormContext extends MinkContext implements SnippetAccepting
             'css',
             sprintf('div.ezfield-identifier-%s fieldset input', self::$fieldIdentifier)
         );
-
         Assertion::assertNotEmpty($inputNodeElements, 'The input field is not marked as required');
-
         foreach ($inputNodeElements as $inputNodeElement) {
             Assertion::assertEquals(
                 'required',
@@ -162,5 +167,19 @@ final class FieldTypeFormContext extends MinkContext implements SnippetAccepting
                 )
             );
         }
+    }
+
+    /**
+     * Set a field definition option $option to $value.
+     *
+     * @param $option string The field definition option
+     * @param $value mixed The option value
+     */
+    public function setFieldDefinitionOption($option, $value)
+    {
+        $this->contentTypeContext->updateFieldDefinition(
+            self::$fieldIdentifier,
+            new FieldDefinitionUpdateStruct(['fieldSettings' => [$option => $value]])
+        );
     }
 }
--- a/features/FieldTypeForm/selection_fieldtype_edit_form.feature
+++ b/features/FieldTypeForm/selection_fieldtype_edit_form.feature
@@ -1,0 +1,69 @@
+Feature: Selection field value edit form
+    In order to edit content of a Selection field
+    As an integrator
+    I want the Selection field form to implement the FieldType's behaviour
+
+Background:
+    Given a Content Type with an ezselection field definition
+
+Scenario: The attributes of the field have a form representation
+    Given I view the edit form for this field
+     Then the edit form should contain a fieldset named after the field definition
+      And it should contain a select field
+
+Scenario: The options added to a field definition have a form representation
+    Given I add the following options to the field definition:
+          | one |
+          | two |
+          | three |
+     When I view the edit form for this field
+     Then the form element should contain the following values:
+          | one |
+          | two |
+          | three |
+
+Scenario: The select field is flagged as required when the field definition is required and set to multiple choice
+    Given the field definition is required
+      And the field definition is set to multiple choice
+     When I view the edit form for this field
+     Then the select field should be flagged as required
+
+Scenario: The multiple/single checkbox is honored
+    Given the field definition is set to single choice
+     When I view the edit form for this field
+     Then the input is a single selection dropdown
+     When the field definition is set to multiple choice
+      And I view the edit form for this field
+     Then the input is a multiple selection dropdown
+
+Scenario: The first item of the required select field is selected by default when editing
+    Given I add some items to the field definition
+      And the field definition is required
+      And the field definition is set to single choice
+     When I view the edit form for this field
+     Then the first item in the list is selected
+    Given the field definition is set to multiple choice
+     When I view the edit form for this field
+     Then the first item in the list is selected
+
+Scenario: The selected item(s) are correctly saved
+    Given I add the following options to the field definition:
+          | uno |
+          | dos |
+          | tres |
+      And the field definition is set to single choice
+      And I view the edit form for this field
+      And I select the item "dos"
+     When I submit the form
+      And I view the Content that was created
+     Then the field value is "dos"
+    Given the field definition is set to multiple choice
+      And I view the edit form for this field
+      And I select these items:
+          | uno |
+          | tres |
+     When I submit the form
+      And I view the Content that was created
+     Then the field value is:
+          | uno |
+          | tres |
--- a/behat_suites.yml
+++ b/behat_suites.yml
@@ -12,5 +12,7 @@ repository-forms:
             paths:
                 - vendor/ezsystems/repository-forms/features/FieldTypeForm
             contexts:
-                - EzSystems\RepositoryForms\Features\Context\FieldTypeFormContext
                 - EzSystems\RepositoryForms\Features\Context\ContentType
+                - EzSystems\RepositoryForms\Features\Context\FieldTypeFormContext
+                - EzSystems\RepositoryForms\Features\Context\SelectionFieldTypeFormContext
+                - Behat\MinkExtension\Context\MinkContext
--- a/features/Context/ContentType.php
+++ b/features/Context/ContentType.php
@@ -108,5 +108,9 @@ final class ContentType extends RawMinkContext implements Context, SnippetAccept
         );
 
         $this->contentTypeService->publishContentTypeDraft($contentTypeDraft);
+
+        $this->currentContentType = $this->contentTypeService->loadContentTypeByIdentifier(
+            $this->currentContentType->identifier
+        );
     }
 }
--- a/features/Context/SelectionFieldTypeFormContext.php
+++ b/features/Context/SelectionFieldTypeFormContext.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\Features\Context;
+
+use Behat\Behat\Context\SnippetAcceptingContext;
+use Behat\Behat\Hook\Scope\BeforeScenarioScope;
+use Behat\MinkExtension\Context\RawMinkContext;
+use PHPUnit_Framework_Assert as Assertion;
+
+final class SelectionFieldTypeFormContext extends RawMinkContext implements SnippetAcceptingContext
+{
+    private static $fieldIdentifier = 'field';
+
+    /**
+     * @var \EzSystems\RepositoryForms\Features\Context\FieldTypeFormContext
+     */
+    private $fieldTypeFormContext;
+
+    /** @BeforeScenario */
+    public function gatherContexts(BeforeScenarioScope $scope)
+    {
+        $this->fieldTypeFormContext = $scope->getEnvironment()->getContext(
+            'EzSystems\RepositoryForms\Features\Context\FieldTypeFormContext'
+        );
+    }
+
+    /**
+     * @Given /^the field definition is set to single choice$/
+     */
+    public function setFieldDefinitionToSingleChoice()
+    {
+        $this->fieldTypeFormContext->setFieldDefinitionOption('isMultiple', false);
+    }
+
+    /**
+     * @Given /^the field definition is set to multiple choice$/
+     */
+    public function setFieldDefinitionToMultipleChoice()
+    {
+        $this->fieldTypeFormContext->setFieldDefinitionOption('isMultiple', true);
+    }
+
+    /**
+     * @Then it should contain a select field
+     */
+    public function itShouldContainASelectField()
+    {
+        $this->assertSession()->elementExists(
+            'css',
+            sprintf(
+                'div.ezfield-identifier-%s fieldset select',
+                self::$fieldIdentifier
+            )
+        );
+    }
+
+    /**
+     * @Then the select field should be flagged as required
+     */
+    public function theSelectFieldShouldBeFlaggedAsRequired()
+    {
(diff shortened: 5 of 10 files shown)


bdunogier:
Detail: no need for "fix", it is not an issue :-)

That aside, I have tested the branch, and I'm getting an array to string conversion error if the selection isn't a multiple one. It works if it is.

glye:
Weird, that's precisely the problem I had before adding the transformer, was working after that. Will retest.

glye:
Ah, a silent error. The earlier one certainly wasn't. Fixed :)

glye:
Test done. Sorry for the delay, I misunderstood how it was supposed to deal with null cases. Should be good now.

glye:
Added BDD (I think). Review ping @bdunogier @yannickroger

andrerom:
Side: One of the Form related BDD scenarios should be star wars centric and end with *May the Form be with you.*

May the Force be with you

glye:
> should be star wars centric

> The Travis CI build failed

I have a bad feeling about this.

glye:
@bdunogier Symfony doesn't set the "required" attribute when in single select mode, it just removes the empty default option. That has the same effect for the end user, but I need to adapt the BDD to that.

bdunogier:
About BDD and select, I see two options.

A. Use a custom sentence for this case, that describes the expectation
B. Enhance the current sentence's implementation to cover that case.

I'd prefer B from where I'm standing.

glye:
Actually, it *also* fails because a select is not an input, it's a select :) So with that it looks more like custom sentences to me. I think I grok it now.

bdunogier:
Well, select is kind of an input, _somehow_.

features/FieldTypeForm/selection_fieldtype_edit_form.feature
```diff
@@ -0,0 +1,22 @@
...
+    Given a Content Type with an ezselection field definition
+
+Scenario: The attributes of the field have a form representation when the field definition is set to multiple choice
+    When I view the edit form for this field
+    Then the edit form should contain a fieldset named after the field definition
+     And it should contain a multi-selection input field
+
+Scenario: The attributes of the field have a form representation when the field definition is not set to multiple choice
+    When I view the edit form for this field
+    Then the edit form should contain a fieldset named after the field definition
```
bdunogier: I would not repeat that, it is already covered above.

glye:
Update: A little more BDD. Not passing yet. Curious to see how it goes on travis. Where the test expects `div.ezfield-identifier-myfield` locally I get `div.ez-field--myfield`.

glye:
Damn, I didn't expect the previous push to pass.
This is it, in theory, but waiting for travis since none of this is passing locally, some problem with my setup still.

features/Context/FieldTypeFormContext.php
```diff
@@ -163,4 +164,118 @@ public function theInputFieldsShouldBeFlaggedAsRequired()
...
+        $this->assertElementOnPage(
+            sprintf(
+                'div.ezfield-identifier-%s fieldset select',
+                self::$fieldIdentifier
+            )
+        );
+    }
+
+    /**
+     * @Given the field definition is not set to multiple choice
```
bdunogier: Is set to single choice ?

glye: It's a "multiple" checkbox, yes. Otherwise I'd have used "single choice". Thanks for the clarification.

features/FieldTypeForm/selection_fieldtype_edit_form.feature
```diff
@@ -0,0 +1,25 @@
...
+      And the field definition is set to multiple choice
+     When I view the edit form for this field
+     Then the select field should be flagged as required
+
+Scenario: The multiple/single checkbox is honored
+    Given that I disable the field definition "isMultiple" option
+     When I view the edit form for this field
+     Then the input is a single selection dropdown
+     When I enable the field definition "isMultiple" option
+     Then the input is a multiple selection dropdown
```
bdunogier: It isn't, unless you view the edit form again... :-)

```
      When I enable the field definition "isMultiple" option
+      And I view the edit form for this field
      Then the input is a multiple selection dropdown
```

glye:
For the "required" check, trying hasElement rather than getElement, as the "required" attribute may not necessarily have a value.

glye:
Desperate hack to find out what HTML travis sees, because it's not the same as for my local behat.

bdunogier:
It looks like it has `"multiple"` alone, not `multiple="multiple"`. Odd...

glye:
Yep, but that at least works, if not actually being XHTML compatible. Why the smeg is "required" missing?

bdunogier:
I really can't say, @glye... 

On my machine, both `multiple="multiple"` and `required="required"` show up as expected.

glye:
Fix content type of behat context not being reloaded after updating the field definition. Muchos gracias @bdunogier

glye:
First refactoring to reduce redundancy. Note that these sentences are still redundant: `the field definition is set to multiple choice` and `I enable the field definition "isMultiple" option`. The first is clearer, and the string "isMultiple" is not visible to the end user anyway. @bdunogier if you agree I'll use the first approach in all cases.

The "flagged as required" tests can also be refactored.

glye:
Refactored the "flagged as required" tests.

glye:
@bdunogier Added scenario for testing selection options added to the field type. (Depending on urgency we could postpone the implementation.)
Should we also test here that a selection made by a user is correctly saved?

andrerom:
Another Side: F&T requested support for default selection, given this afaik was never requested on legacy, did it auto select first item?

glye:
@andrerom Legacy autoselects the first item of single selects, not multi selects. Attribute "Required" property doesn't matter.

andrerom:
@glye could you make sure we do the same here? And we'll add enhancement to do it also in Platform UI?

glye:
@andrerom I'm sure it's possible, but we'd have to "fight" symfony to do it - imho the symfony behaviour makes more sense:
- Single select, required: Autoselection could make sense here
- Single select, not required: Symfony inserts an empty first item in the dropdown (this can be disabled), to enable a zero selection. This makes sense imho. We *could* still autoselect the first "real" item, but it sounds... not right.
- Multiselect, required: Autoselection could make sense here, though legacy doesn't do it.
- Multiselect, not required: Autoselection makes little sense, and legacy doesn't do it.

What do you think?
Did they request autoselection also for multiselects?

bdunogier:
I agree with `the field definition is set to multiple choice` over the other one. The implementation could call a more generic method that works for all flags.

I'm just thinking that it might make more sense to move sentences dedicated to particular fieldtypes to dedicated contexts, and keep the FieldTypeContext clean ?

bdunogier:
> Should we also test here that a selection made by a user is correctly saved?

We will have to yes. 

> Single select, not required: Symfony inserts an empty first item in the dropdown (this can be disabled), to enable a zero selection. This makes sense imho. We could still autoselect the first "real" item, but it sounds... not right.

I agree with that. Why would we force-select the first item if the field isn't required ?

> Multiselect, required: Autoselection could make sense here, though legacy doesn't do it.

It would make sense I guess, since it is what will happen with a single select. The difference is that with a single-select, since there isn't an empty choice if it is required, something _has_ to be selected, while a multi-select can exist without an option selected. 

So maybe I'd lean towards auto-select in both cases. But I feel like we're gonna be asked to make the default choice configurable...

andrerom:
I agree, only auto select on `Single select, required: Auto selection could make sense here` where it makes sense in symfony. Lets not fight it :)

> So maybe I'd lean towards auto-select in both cases. But I feel like we're gonna be asked to make the default choice configurable...

I'm a bit torn here, but I also lean towards selecting first one here as well, as it feels more consistent.

---

As we will need to do this in Platform UI as well, @dpobel / @rolandbenedetti Any input here?

TL;DR; This is to solve "default selection" use case request from FT:
- We will auto select first item if single selection, and field is required  - Same as legacy
- We will auto select first item if multi selection, and field is required  - Legacy does not have this, but feels most consistent.

glye:
@bdunogier @andrerom Added BDD features for autoselection and storing content, in https://github.com/ezsystems/repository-forms/pull/84/commits/8bdeafa46daa384d1044e38e25898e63b665d96b

dpobel:
> TL;DR; This is to solve "default selection" use case request from FT:

>    We will auto select first item if single selection, and field is required - Same as legacy

Sure of that ? I quickly looked at how Legacy handles the selection datatype and I see no trace of that. But it's the default behavior of the HTML `select` in the browser, if no option is set as selected, the first one appears as visible/selected even if it's not explicitly selected with the corresponding HTML attribute, so normally it's done for free (unless the Form component or something adds an empty option in the beginning)

>    We will auto select first item if multi selection, and field is required - Legacy does not have this, but feels most consistent.

well maybe it's more consistent but the default multiple select UI in browser is just horrible. Very few people knows that you can select several options by using Control :) That's one of the reason why in PlatformUI, we have a special widget for that and we don't rely on a `<select multiple>`. Simple way to improve the UI by just using some simple HTML is to replace the `<select multiple>` by an unordered list containing one checkbox for each option. That said, I'm unsure what this involves in terms of Form component integration :)

glye:
@dpobel 
> That's one of the reason why in PlatformUI, we have a special widget for that and we don't rely on a <select multiple>.

Yeah, I am curious how that will work regarding autoselection.

> Simple way to improve the UI by just using some simple HTML is to replace the <select multiple> by an unordered list containing one checkbox for each option. That said, I'm unsure what this involves in terms of Form component integration :)

Symfony-wise that is dead simple, it's just configuration of the form field. Same to get radio buttons for a single select.

glye:
To combat PR growth I have moved out these bits:
- Autoselect first item: https://jira.ez.no/browse/EZP-25851
- Implement BDD sentences that were added late: https://jira.ez.no/browse/EZP-25852

The rest is ready for review, ping @bdunogier @andrerom

features/Context/FieldTypeFormContext.php
```diff
@@ -144,23 +145,120 @@ public function theFieldDefinitionIsMarkedAsRequired()
      */
     public function theInputFieldsShouldBeFlaggedAsRequired()
     {
-        $inputNodeElements = $this->getSession()->getPage()->findAll(
+        $this->theFieldShouldBeFlaggedAsRequired('input');
+    }
+
+    /**
+     * @Given /^the field definition is set to single choice$/
```
bdunogier: You could go for `/^the field definition is set to :value choice$/`, with a method argument, and handle single/multiple in the method. It would allow you to remove one. Just unsure if it is worth it or not :)

features/Context/FieldTypeFormContext.php
```diff
@@ -144,23 +145,120 @@ public function theFieldDefinitionIsMarkedAsRequired()
...
         }
     }
+
+    /**
+     * Set a boolean field definition option $option to $value.
+     *
+     * @param $option string The field definition option
+     * @param $value bool The option value
+     */
+    protected function setBooleanFieldDefinitionOption($option, $value)
```
bdunogier: Is it limited to boolean fields ? It would work with any field with the current code, would it not ?

I'd also make this method public, since it can be used by other contexts. I have drafted an example we might want to use:

```
<?php
namespace EzSystems\RepositoryForms\Features\Context;

use Behat\Behat\Context\SnippetAcceptingContext;
use Behat\Behat\Hook\Scope\BeforeScenarioScope;
use Behat\MinkExtension\Context\RawMinkContext;

class SelectionFieldTypeFormContext extends RawMinkContext implements SnippetAcceptingContext
{
    /**
     * @var \EzSystems\RepositoryForms\Features\Context\FieldTypeFormContext
     */
    private $fieldTypeFormContext;

    /** @BeforeScenario */
    public function gatherContexts(BeforeScenarioScope $scope)
    {
        $this->fieldTypeFormContext = $scope->getEnvironment()->getContext(
            'EzSystems\RepositoryForms\Features\Context\FieldTypeFormContext'
        );
    }

    /**
     * @Given /^the field definition is set to single choice$/
     */
    public function setFieldDefinitionToSingleChoice()
    {
        $this->fieldTypeFormContext->setFieldDefinitionOption('isMultiple', false);
    }

    /**
     * @Given /^the field definition is set to multiple choice$/
     */
    public function setFieldDefinitionToMultipleChoice()
    {
        $this->fieldTypeFormContext->setFieldDefinitionOption('isMultiple', true);
    }
}
```

(It does imply a renaming of the method to `setFieldDefinitionOption`, but the idea is the same)

features/Context/FieldTypeFormContext.php
```diff
@@ -144,23 +145,120 @@ public function theFieldDefinitionIsMarkedAsRequired()
...
+                self::$fieldIdentifier
+            ),
+            'multiple="multiple"'
+        );
+    }
+
+    /**
+     * @Then the input is a multiple selection dropdown
+     */
+    public function theInputIsAMultipleSelectionDropdown()
```
bdunogier: If we decide to use dedicated context for dedicated FieldTypes, maybe these should go to the `SelectFieldTypeFormContexts` ?

features/FieldTypeForm/selection_fieldtype_edit_form.feature
```diff
@@ -0,0 +1,72 @@
...
+     Then the edit form should contain a fieldset named after the field definition
+      And it should contain a select field
+
+Scenario: The options added to a field definition have a form representation
+    Given I add the following options to the field definition:
+          | one |
+          | two |
+          | three |
+     When I view the edit form for this field
+     Then the following options should be represented in the form:
```
bdunogier: Should we make this a bit more targetted at the expected behaviour ? Something about the dropdown: "The the dropdown should contain the following values:" ?

features/FieldTypeForm/selection_fieldtype_edit_form.feature
```diff
@@ -0,0 +1,72 @@
...
+Scenario: The multiple/single checkbox is honored
+    Given the field definition is set to single choice
+     When I view the edit form for this field
+     Then the input is a single selection dropdown
+     When the field definition is set to multiple choice
+      And I view the edit form for this field
+     Then the input is a multiple selection dropdown
+
+Scenario: The first item of the required select field is selected by default when editing
+    Given I add the following options to the field definition:
```
bdunogier: I'm not sure I'd use an actual value list in the scenario. Would it better describe the behaviour if it ended up with something like "Then the first value in the list is selected by default" ?

lib/FieldType/DataTransformer/SelectionValueTransformer.php
```diff
@@ -0,0 +1,47 @@
...
+
+use eZ\Publish\Core\FieldType\Selection\Value;
+use Symfony\Component\Form\DataTransformerInterface;
+
+/**
+ * DataTransformer for Selection\Value.
+ * Needed to display the form field correctly and transform it back to an appropriate value object,
+ * as single selection form fields return integer values, while multiselects return arrays.
+ */
+class SelectionValueTransformer implements DataTransformerInterface
```
bdunogier: By reading the description, I'm thinking of an alternative: use different data transformers for single and multiple choices. Since we create the transformer from the mapper, we could decide which one to use.

features/FieldTypeForm/selection_fieldtype_edit_form.feature
glye: By dropdown, do you mean the JS widget in PlatformUI? The BDD of repoforms itself uses plain html, doesn't it? So then we may need to cover both dropdown and multiselect cases.
2nd complication: In the current PlatformUI, the items in the select are NOT in fact represented in the form, not until you start typing anyway. Can we make BDD sentences that make sense both in plain html and in PlatformUI?

features/FieldTypeForm/selection_fieldtype_edit_form.feature
glye: I thought a value list was more kosher than hardcoding something in the implementation. If not a list, then something like `Given I add some options to the field definition`?

lib/FieldType/DataTransformer/SelectionValueTransformer.php
glye: Agree, but for a different reason :) I don't see how the unit test as is could get any simpler - I think it might be too simple actually. Single select is not properly covered. And adding that would be a case for separate transformers.

bdunogier: Then it is pretty much the same reason: make it testable ;)

Go for it if you think it is useful.

features/FieldTypeForm/selection_fieldtype_edit_form.feature
glye: Went with `the form element should contain the following values` to be flexible.

features/FieldTypeForm/selection_fieldtype_edit_form.feature
bdunogier: Well, value are, as I see it, more kosher when they impact the behaviour. Here, the values being "one", "two", "three" or "pony", "unicorn", "troll" has absolutely no impact on the behaviour.

For the scenario where we describe that every value added to the field def must appear in the select, then yes, the values matter, as they make that clear to the scenario's reader (even though we could get away with it by phrasing that "Given a set of values added to the field definition... Then the select contains the values from the field definition").

features/Context/FieldTypeFormContext.php
glye: Well, then we remove one method and add two :) Though the new ones might be protected. I'm... https://xkcd.com/1378/

features/FieldTypeForm/selection_fieldtype_edit_form.feature
glye: Then I see no difference between the two scenarios. It doesn't matter what the values are, as long as output matches input. Whichever you prefer :)

features/Context/FieldTypeFormContext.php
glye: Ok, looks good.

features/Context/FieldTypeFormContext.php
bdunogier: No problem. You've quoted my favourite XKCD comic, I can't argue against that :-)

features/FieldTypeForm/selection_fieldtype_edit_form.feature
bdunogier: As I tried to phrase, when the scenario is about making sure that configured values show up in the input field, listing them is I think okay.

But about the default selection, I think we can do without the values, and phrase what the expected behaviour actually is. The way it is phrased right now, maybe it will select by default the value that starts with 'u', or the one that matches the name of a game ;)

bdunogier:
+1 once the last batch of BDD improvements is taken care of.

glye:
Large pile of changes, need to let Travis have a look at it since BDD not fully working locally. Also squashed the lengthy history.
- Separate transformers/tests for single/multi select https://github.com/ezsystems/repository-forms/pull/84/commits/b3fbd3cbc609151441329baced00bcf87d1c0cd0
- BDD sentence changes https://github.com/ezsystems/repository-forms/pull/84/commits/061e77df7d719a8563540edda9dced0c69283929
- Separate context for the Selection type https://github.com/ezsystems/repository-forms/pull/84/commits/53cc89d25cd302ccc1664721dddcaf44e3ad5ca1

glye:
@bdunogier When `SelectionFieldTypeFormContext` extends `RawMinkContext` behat fails because `assertElementOnPage()` isn't defined. When it extends `MinkContext`, it fails because `iAmOnHomepage()` is defined twice. You know the right way out of this Catch-22?

bdunogier:
You can't use in the same suite two or more Contexts that implement MinkContext, you got that right.

If all you need is `assertElementOnPage()`, use `RawMinkContext` (as you did figure out), and use `$this->assertSession()->elementExists()` (that's probably what `MinkContext` does (checked: precisely)).

glye:
Update: Using `RawMinkContext`, and updated element assertions.

glye:
@bdunogier Passing now (with some not-yet-implemented scenarios, ref https://jira.ez.no/browse/EZP-25852 )

features/FieldTypeForm/selection_fieldtype_edit_form.feature
```diff
@@ -0,0 +1,69 @@
...
+      And I view the edit form for this field
+     Then the input is a multiple selection dropdown
+
+Scenario: The first item of the required select field is selected by default when editing
+    Given I add some items to the field definition
+      And the field definition is required
+      And the field definition is set to single choice
+     When I view the edit form for this field
+     Then the first item in the list is selected
+     When the field definition is set to multiple choice
```
bdunogier: This would be a Given, imho.

features/FieldTypeForm/selection_fieldtype_edit_form.feature
```diff
@@ -0,0 +1,69 @@
...
+          | uno |
+          | dos |
+          | tres |
+      And the field definition is set to single choice
+      And I view the edit form for this field
+      And I select the item "dos"
+     When I submit the form
+      And I view the Content that was created
+     Then the field value is "dos"
+     When the field definition is set to multiple choice
```
bdunogier: Also a given.

bdunogier:
Looks good, besides the two very minor remarks.

+1

glye:
Fixed the Givens in the multi-part scenarii.

bdunogier:
Looks good. I think you can even squash everything upon merge.

PR #86, opened 2016-06-01
Added pagelayout to content_edit template/controller

bdunogier:
Title pretty much says it all. Ensures that UGC forms are well integrated in the site.

### Questions
- [ ] See if/how we want to cover this with BDD. Sentence is easy, but what about the actual assertion(s) ?

### DIFF
--- a/behat_suites.yml
+++ b/behat_suites.yml
@@ -8,7 +8,7 @@ repository-forms:
             contexts:
                 - EzSystems\RepositoryForms\Features\Context\ContentEdit
                 - EzSystems\RepositoryForms\Features\Context\ContentType
-                - EzSystems\RepositoryForms\Features\Context\PageLayoutContext
+                - EzSystems\RepositoryForms\Features\Context\PagelayoutContext
         fieldtype_form:
             paths:
                 - vendor/ezsystems/repository-forms/features/FieldTypeForm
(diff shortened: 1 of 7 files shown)


bdunogier:
Ping @glye

glye:
I second @ezrobot. And it's not pretty on my setup, but I guess that's a) a local installation problem and b) bling is out of scope now anyway.
So it looks like +1 after CS.

bdunogier:
CS fixed :)

glye:
About BDD, there are many assertions one could make, but we want something that doesn't fail when the pagelayout changes. So how about basing it on the state before this PR (no structure around the form) and assert that the surrounding structure is *not nothing*? Would that work? Double negative, I know :)

bdunogier:
I don't really see the point of comparing the new state to before the pull-request. It was in any case, as I see it, a temporary situation.

What we want to verify, I believe, is that the _configured pagelayout_ is used to render the form. The question would be how to assert that exactly. The simplest would be to check that we do have the global (html, body, etc) markup, but it doesn't _really_ assert that the configured pagelayout is used.

glye:
> The simplest would be to check that we do have the global (html, body, etc) markup

This is what I said, I believe. But communication etc. :) Agree on checking configured pagelayout.

bdunogier:
Then we are on the same page, great :)

glye:
Uh, where does the `<!-- STOP %s -->` come from? Do you insert it into the pagelayout template for the purpose of the test, or is it already there?

bdunogier:
The comment thing is there when the kernel has debugging enabled (the behat & dev env do). I initially wanted to use the START one but it isn't in the HTML captured by Mink for some reason (probably because it appears _before_ the `<html>` tag.

glye:
+1

andrerom:
+1

PR #87, opened 2016-06-03
User registration support (EZP-25730)

bdunogier:
> Implements [EZP-25730|https://jira.ez.no/browse/EZP-25730]

Adds real user registration support through the `/register` route. Add the `user/register` policy to the anonymous user to enable it (or use the admin account). After registering, the user is redirected to a page that confirms registration.

user_register_screenshot

### TODO
- [x] Add a custom `user/register` policy, and deny the `/register` route if the user does not have this policy
- [ ] Make target user group configurable (#88)
- [ ] Make the identifier of the Content Type used for user registration configurable
- [ ] Read user creation language from prioritized languages
- [ ] Fix the register button's label
- [ ] BDD coverage
- [ ] Improve the user account's form formatting ?

### DIFF
--- a/bundle/Controller/UserRegisterController.php
+++ b/bundle/Controller/UserRegisterController.php
@@ -12,6 +12,7 @@ use eZ\Bundle\EzPublishCoreBundle\Controller;
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\Repository;
 use eZ\Publish\API\Repository\UserService;
+use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
 use EzSystems\RepositoryForms\Data\Mapper\UserCreateMapper;
 use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\User\UserRegisterType;
@@ -73,9 +74,15 @@ class UserRegisterController extends Controller
      * @param Request $request
      *
      * @return \Symfony\Component\HttpFoundation\Response
+     *
+     * @throws \Exception if the current user isn't allowed to register an account
      */
     public function registerAction(Request $request)
     {
+        if (!$this->isGranted(new Attribute('user', 'register'))) {
+            throw new \Exception('You are not allowed to register a new account');
+        }
+
         $language = 'eng-GB';
 
         $contentType = $this->repository->sudo(
--- a/bundle/Controller/UserRegisterController.php
+++ b/bundle/Controller/UserRegisterController.php
@@ -11,11 +11,11 @@ namespace EzSystems\RepositoryFormsBundle\Controller;
 use eZ\Bundle\EzPublishCoreBundle\Controller;
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\Repository;
-use eZ\Publish\API\Repository\UserService;
 use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
 use EzSystems\RepositoryForms\Data\Mapper\UserRegisterMapper;
 use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\User\UserRegisterType;
+use EzSystems\RepositoryForms\UserRegister\RegistrationGroupLoader;
 use Symfony\Component\HttpFoundation\Request;
 
 class UserRegisterController extends Controller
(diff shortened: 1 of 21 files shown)


bundle/Controller/UserRegisterController.php
```diff
@@ -0,0 +1,126 @@
...
+    {
+        if (!$this->isGranted(new Attribute('user', 'register'))) {
+            throw new \Exception('You are not allowed to register a new account');
+        }
+
+        $language = 'eng-GB';
+
+        $contentType = $this->repository->sudo(
+            function () {
+                return $this->contentTypeService->loadContentTypeByIdentifier('user');
```
lolautruche: The identifier should be configurable, like the user group

bundle/Controller/UserRegisterController.php
```diff
@@ -0,0 +1,126 @@
...
+            $this->contentActionDispatcher->dispatchFormAction($form, $data, $form->getClickedButton()->getName());
+            if ($response = $this->contentActionDispatcher->getResponse()) {
+                return $response;
+            }
+        }
+
+        return $this->render('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig', [
+            'form' => $form->createView(),
+            'languageCode' => $language,
+            'pagelayout' => $this->pagelayout,
```
lolautruche: Can be null if not set by the `setPagelayout()` setter. Is this wanted?

lib/Form/Processor/UserRegisterFormProcessor.php
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Saves content draft corresponding to $data.
+     * Depending on the nature of $data (create or update data), the draft will either be created or simply updated.
+     *
+     * @param ContentStruct|\EzSystems\RepositoryForms\Data\User\UserCreateData $data
+     * @param $languageCode
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Content
+     */
+    private function saveDraft(UserCreateData $data, $languageCode)
```
lolautruche: Where is this method called?

lib/Form/Type/User/UserRegisterType.php
```diff
@@ -0,0 +1,68 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Form type for content edition (create/update).
+ * Underlying data will be either \EzSystems\RepositoryForms\Data\Content\ContentCreateData or \EzSystems\RepositoryForms\Data\Content\ContentUpdateData
+ * depending on the context (create or update).
+ */
+class UserRegisterType extends AbstractType
+{
+    public function getName()
```
lolautruche: Implementing this method is deprecated

lib/Form/Processor/UserRegisterFormProcessor.php
bdunogier: Line 64 above ?  :)

bundle/Controller/UserRegisterController.php
bdunogier: Shouldn't be a problem, the templates checks it.

Otoh, maybe it doesn't make much sense not to set it.

bundle/Controller/UserRegisterController.php
bdunogier: Todo item added.

lib/Form/Type/User/UserRegisterType.php
bdunogier: I can indeed see that, but the `FormTypeInterface` still contains it. I don't see how I could not implement it without getting myself a fatal error :)

glye:
Re: your "not a huge fan", it is a bit weird to read. I was expecting `iDo(Not)HaveTheUserRegisterPolicy()` to login and then check the policy, but I guess the point here is not the api for reading roles, but the register action, so it does the job with a bit less code. Also re: earlier discussion about actions that change things vs assertions :) The private methods seem okay.

bdunogier:
> iDo(Not)HaveTheUserRegisterPolicy() to login and then check the policy, but I guess the point here is not the api for reading roles

yeah, that was what I was imagining. But we don't have a way to check that using mink (we probably could with REST). Then after thinking, I concluded that it was _much_ easier that way.

About assertions/changes, it feels alright in that case, but it is a matter of perspective I guess. It is clear here that for the policies scenarii, we need to customize roles, we can't just verify, as we do not have a role with that policy in the default data. We could use the normal anonymous role for the negative check, but it will fail when we update data to allow registration by default.

Overall, creating a user with a particular role is something we do quite often. It is a bit odd to do it for the anonymous one, but our repository permits it :-)

glye:
> scenarii

:tophat:

> In any case, i'd rather keep creating custom roles for this

To clear that up, I have nothing against creating the role. It's just that given the sentence/method naming, I was at first expecting the role to be checked at some point. Given that a) it *is* created and b) we don't care about role reading api here, it's ok :)

+1

bdunogier:
Closing to reopen against master.

PR #88, opened 2016-06-06
Made user registration group configurable

bdunogier:
> Implements [EZP-25840](https://jira.ez.no/browse/EZP-25840)

Adds a `RegistrationGroupLoader`, with a `ConfigurableRegistrationGroupLoader` that receives a newly added `user.registration_group_id` siteaccess aware configuration variable. This variable is configured as follows:

```yaml
ezpublish:
  system:
    <scope>:
      users:
        registration_group_id: 11
```

### DIFF
--- a/bundle/DependencyInjection/Configuration/Parser/UserRegistration.php
+++ b/bundle/DependencyInjection/Configuration/Parser/UserRegistration.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryFormsBundle\DependencyInjection\Configuration\Parser;
+
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\AbstractParser;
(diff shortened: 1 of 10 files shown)


bundle/DependencyInjection/Configuration/Parser/UserRegistration.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public function addSemanticConfig(NodeBuilder $nodeBuilder)
+    {
+        $nodeBuilder
+            ->arrayNode('users')
+                ->info('Users configuration')
+                ->children()
+                    ->scalarNode('registration_group_id')
+                        ->info('Content id of the user group where users who register are created.')
+                        ->defaultValue(11)
```
glye: This will be doc-ed somewhere, I presume?
Also, will you add the setting to platform repo config files, where people will discover it on their own?

glye:
Dang. This was rather easier in legacy - perhaps that's why it had so many settings :sweat_smile:
+1

bundle/DependencyInjection/Configuration/Parser/UserRegistration.php
bdunogier: > This will be doc-ed somewhere, I presume?

Well, it is already documented when running `debug:configuration ezpublish` ;)
But yes, it will require its own entry in the doc, indeed.

> Also, will you add the setting to platform repo config files, where people will discover it on their own?

I didn't really intend to do that, but we could, no strong opinion.
Maybe @andrerom has an opinion about that ?

PR #90, opened 2016-06-08
EZP-25863: Made user register use the main language

bdunogier:
> Implements [EZP-25863](https://jira.ez.no/browse/EZP-25863)

Changes the hardcoded `"eng-GB"` to the site's most prioritized language.

The `UserRegisterMapper` is now a service. It is changed to accept parameters using a `setParam` method, that accepts `"language"`. The language is read from the configuration using the expression language (see `services.yml`).

Other changes (refactoring commit):
- ContentType and Parent Group loading is now handled by dedicated loaders ( `RegistrationContentTypeLoader` and `RegistrationGroupLoader`), both used by the `UserRegisterMapper`.

### TODO
- [x] Split the Content Type change to its own commit.
- [ ] Tests

### DIFF
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -49,6 +49,8 @@ parameters:
 
     ezrepoforms.controller.content_edit.class: EzSystems\RepositoryFormsBundle\Controller\ContentEditController
 
+    ezrepoforms.user_content_type_identifier: "user"
+
 services:
     # deprecated in 1.1, will be removed in 2.0
     ezrepoforms.field_type_form_mapper.registry:
@@ -301,7 +303,6 @@ services:
     ezrepoforms.controller.user_register:
         class: 'EzSystems\RepositoryFormsBundle\Controller\UserRegisterController'
         arguments:
-            - "@ezpublish.api.service.content_type"
             - "@ezrepoforms.form_data_mapper.user_register"
             - "@ezrepoforms.user_register.registration_group_loader.configurable"
             - "@ezrepoforms.action_dispatcher.content"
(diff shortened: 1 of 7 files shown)


andrerom:
does this have an impact on the labels as well? *(from contentType field names)*
Cause on them I would expect the code to use one of the available twig helpers to pick by prioritised language.

bdunogier:
> does this have an impact on the labels as well? (from contentType field names)

It might, I need to check.

bundle/Resources/config/services.yml
```diff
@@ -315,6 +317,14 @@ services:
...
         arguments:
-            - @ezpublish.api.repository
+            - "@ezpublish.api.repository"
         calls:
             - [setGroupId, ["$users.registration_group$"]]
+
+    ezrepoforms.form_data_mapper.user_register:
+        class: EzSystems\RepositoryForms\Data\Mapper\UserRegisterMapper
+        arguments:
+            - "@ezpublish.api.repository"
```
glye: Why the quotes here? Same for `registration_group_loader.configurable` above.

crevillo: very good one @glye. 
it seems YAML specification doesn't allow start northing with `@`. so you MUST quote it. It happens symfony 2 yaml parser allows to do that and this is cause it's working for ez until now. 

But looks symfony 3 will follow that YAML spec so maybe we will need to quote all this kind of service arguments...

glye: Aha, that would be a good reason :) Thanks.

glye:
Seems OK to me, fwiw.

bundle/Resources/config/services.yml
bdunogier: Reason is Sf 3.0 compatibility. We will start getting deprecation notices unless we add those quotes. You should actually do the same from now on :-)

bdunogier:
I have changed pretty much everything. Please have a look again :)

glye:
So the use of `sudo()` here isn't new, only the abstraction is, afaict. OK. Does `ConfigurableSudoRepositoryLoader` really belong under `UserRegister`? It sounds more generic.

bdunogier:
> Does ConfigurableSudoRepositoryLoader really belong under UserRegister? It sounds more generic.

I agree. On the other hand, I'm not really sure _right now_ where it should go (could even be kernel). I think it's fine to put it there, and deprecate it when/if we find a better home for it.

glye:
Sure, fine with that.

andrerom:
> Does ConfigurableSudoRepositoryLoader really belong under UserRegister? It sounds more generic.

be aware, if sudo dies at some point, it will rather be about loading the repository that does not check permissions. so it'a like a root repository or whatever. naming.

bdunogier:
> be aware, if sudo dies at some point, it will rather be about loading the lower level repository instance that does not check permissions. so it'a like a root repository or whatever. naming.

In that case, we will deprecate the "sudo" implementation, and add a new one that uses the RootRepository or whatever ?

lib/UserRegister/ConfigurableSudoRepositoryLoader.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @return Repository
+     */
+    protected function getRepository()
+    {
+        return $this->repository;
+    }
+
+    protected function load(Closure $callback)
```
andrerom: load is a bit misleading. could just as well be called sudo.

bdunogier: will rename

bdunogier:
Comments handled. Will merge to feature when travis agrees.

PR #91, opened 2016-06-15
Make user registration templates customizable

bdunogier:
> Implements [EZP-25373](https://jira.ez.no/browse/EZP-25373)
> Status: ready for review
> [Behat scenario](https://github.com/ezsystems/repository-forms/blob/ezp25373-user_registration_templates/features/User/Registration/user_registration.feature#L36-L74)

### Usage
Registration form and confirmation templates can be customized using the `ezpublish.user_registration.templates` siteaccess aware settings:

```twig
ezpublish:
  system:
    default:
      user_registration:
        templates:
          form: 'path/to/user/registration_form.html.twig'
          confirmation: 'path/to/user/registration_confirmation.html.twig'
```

Registration group configuration is adapted to the `users` -> `user_registration` setting modification above:

```
ezpublish:
  system:
    default:
      user_registration:
        group_id: 123
```

### Internal changes
- Changes the UserController to return `View` objects (`UserRegisterFormView` and `UserRegisterConfirmView`)
- Adds `$user_registration.templates.form$` and `$user_registration.templates.confirmation$` siteaccess aware container parameters
- Adds a `UserRegistrationViewTemplates` event listener that receives the configured templates and, when applicable, sets them in the View objects when applicable
- Removes `pagelayout` handling in the controller, and moves it to the event listener

### TODO
- [x] Fix the failures
- [x] Templates semantic config

### DIFF
--- a/bundle/DependencyInjection/Configuration/Parser/UserRegistration.php
+++ b/bundle/DependencyInjection/Configuration/Parser/UserRegistration.php
@@ -23,12 +23,23 @@ class UserRegistration extends AbstractParser
     public function addSemanticConfig(NodeBuilder $nodeBuilder)
     {
         $nodeBuilder
-            ->arrayNode('users')
-                ->info('Users configuration')
+            ->arrayNode('user_registration')
+                ->info('User registration configuration')
                 ->children()
-                    ->scalarNode('registration_group_id')
+                    ->scalarNode('group_id')
                         ->info('Content id of the user group where users who register are created.')
                         ->defaultValue(11)
+                    ->end()
+                    ->arrayNode('templates')
+                        ->info('User registration templates.')
+                        ->children()
+                            ->scalarNode('form')
+                                ->info('Template to use for registration form rendering.')
+                            ->end()
+                            ->scalarNode('confirmation')
+                                ->info('Template to use for registration confirmation rendering.')
+                            ->end()
+                        ->end()
                     ->end()
                 ->end()
             ->end();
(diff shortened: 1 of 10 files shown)


andrerom:
hmm, wouldn't it also be possible to customize template used by using override system?
*Side: Can people use that in combination with query controller btw, or is that one of the possible downsides of the design you talked about?*

bdunogier:
> hmm, wouldn't it also be possible to customize template used by using override system?

It was my initial idea, but it requires more changes. Furthermore, we don't really have the concept of conditional override in this particular context (which conditions would we override on ?).

That's why I've decided to only use a part of this mechanism.

> Side: Can people use that in combination with query controller btw, or is that one of the possible downsides of the design you talked about?

What does "that" refer to ? This feature ? The event we use here is always available, including for ContentView. It can be used to further customize the result of the QueryController actions, yes.

bdunogier:
If you had a configuration in mind @andrerom, please try to describe it here. It would help.

andrerom:
> It was my initial idea, but it requires more changes. Furthermore, we don't really have the concept of conditional override in this particular context (which conditions would we override on ?).

it would have to be a match all, until we have theme systme in place.
So my slight preference long term is that don't need specific settings for each and every template around, but we can deprecate this later to get there *(theming, ..)*.

features/User/Registration/user_registration.feature
```diff
@@ -32,3 +32,42 @@ Scenario: The user group where registered users are created can be customized
...
      When I register a user account
      Then the user is created in this user group
+
+Scenario: The user registration templates can be customized
+    Given the following user registration templates configuration:
+      """
+      ezpublish:
+        system:
+          default:
+            users:
```
andrerom: user_registration

andrerom:
With the changes you mentioned you'd do (se comment) +1

bdunogier:
Whee, BDD is passing.

bdunogier:
Changes to configuration discussed with @andrerom implemented, see behat changes [here](https://github.com/ezsystems/repository-forms/blob/76359811f89fedaf156697113b7b61bfc2c83cbe/features/User/Registration/user_registration.feature#L30-L31) and [there](https://github.com/ezsystems/repository-forms/blob/76359811f89fedaf156697113b7b61bfc2c83cbe/features/User/Registration/user_registration.feature#L30-L31).

bdunogier:
And rebased, ready to be merged.

bdunogier:
Squashed the feature change into one commit, kept the other small fixes separated.

bundle/DependencyInjection/Configuration/Parser/UserRegistration.php
```diff
@@ -23,22 +23,57 @@ class UserRegistration extends AbstractParser
     public function addSemanticConfig(NodeBuilder $nodeBuilder)
     {
         $nodeBuilder
-            ->arrayNode('users')
-                ->info('Users configuration')
+            ->arrayNode('user_registration')
+                ->info('Users registration configuration')
```
glye: Nitpick: If the "Users" is meant as a plural, it should be "User registrations". (If it's a genitive, it should be "User's".)

bdunogier: `User registration` as a feature it is. Will change, thank you.

glye:
+1 after nitpick. It seems unnecessary to keep the last 2 commits separate, can they be merged into the second commit maybe?

bdunogier:
> It seems unnecessary to keep the last 2 commits separate, can they be merged into the second commit maybe?

You're right, will do.

bdunogier:
Fixed remark and squashed the last two commits into one.

andrerom:
+1

bdunogier:
And merged. Thank you guys.

PR #92, opened 2016-06-29
Throw an UnauthorizedHttpException if not allowed to register

bdunogier:
Changes the exception thrown when the user doesn't have the `user/register` policy from a basic `Exception` to `UnauthorizedHttpException`.

Benefit is that it shows a permission error and not a system one in production mode.


andrerom:
+1

PR #93, opened 2016-08-04
EZP-26128: As a Developer I want deprecation issues resolved, so that application logs are clean

alongosz:
JIRA Story: [EZP-26128](https://jira.ez.no/browse/EZP-26128)

**TODO**:
- [x] Quote service references in Yaml files (3e12b51).
- [x] Quote parameter references in Yaml files (93996e2).
- [x] Make sure tests are run with full error reporting (5ffa8a7).
- [x] Cleanup `services.yml` to keep the code consistent (2a86fb4).
- [x] Replace references of deprecated `FieldTypeFormMapperInterface` with `FieldDefinitionFormMapperInterface` (cb8a97c).
- [x] Replace references of deprecated ez.fieldType.formMapper service tag with `ez.fieldFormMapper.value` or/and  `ez.fieldFormMapper.definition` depending on an interface a mapper implements (c7e291b).
- [x] Replace string references of form field types with fully-qualified type class names (d1ac248).
- [x] Replaced deprecated FormTypeInterface::getName with getBlockPrefix (9d35c16).
- [x] Create missing services for some custom form types defined in `EzSystems\RepositoryForms\Form\Type` namespace to be able to reference them with fully-qualified class name instead of [instantiating them manually](https://github.com/ezsystems/PlatformUIBundle/blob/1.4/Controller/SectionController.php#L84) which is deprecated.

@glye: thanks for help with this :)

### DIFF
--- a/bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
+++ b/bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
@@ -54,7 +54,7 @@ class FieldTypeFormMapperDispatcherPass implements CompilerPassInterface
         $formMappers = $container->findTaggedServiceIds('ez.fieldType.formMapper');
         foreach (array_keys($formMappers) as $serviceId) {
             @trigger_error(
-                'The ez.fieldType.formMapper service tag from $serviceId is deprecated in ezsystems/repository-forms 1.3, ' .
+                "The ez.fieldType.formMapper service tag from $serviceId is deprecated in ezsystems/repository-forms 1.3, " .
                 "and will be removed in version 2.0\n" .
                 'Use ez.fieldFormMapper.value and/or ez.fieldFormMapper.definition instead.',
                 E_USER_DEPRECATED
--- a/lib/Form/Type/Content/ContentEditType.php
+++ b/lib/Form/Type/Content/ContentEditType.php
@@ -4,7 +4,6 @@
  *
  * @copyright Copyright (C) eZ Systems AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
- * @version //autogentag//
  */
 namespace EzSystems\RepositoryForms\Form\Type\Content;
 
@@ -23,6 +22,11 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
 class ContentEditType extends AbstractType
 {
     public function getName()
+    {
+        return $this->getBlockPrefix();
+    }
+
+    public function getBlockPrefix()
     {
         return 'ezrepoforms_content_edit';
     }
--- a/lib/Form/Type/Content/ContentEditType.php
+++ b/lib/Form/Type/Content/ContentEditType.php
@@ -9,6 +9,9 @@
 namespace EzSystems\RepositoryForms\Form\Type\Content;
 
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+use Symfony\Component\Form\Extension\Core\Type\HiddenType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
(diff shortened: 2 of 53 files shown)


alongosz:
Ready for a review, ping @andrerom @glye

andrerom:
+1

glye:
+1. All tests passing, but I hope you also did some quick manual tests - editing some content and some content types - as a precaution.

alongosz:
> (...) I hope you also did some quick manual tests - editing some content and some content types - as a precaution.

I checked editing content but didn't check editing content type. Sth fails there, investigating now.

alongosz:
The cause of errors while editing content type was that `FieldDefinitionType::buildForm` used registry instead of dispatcher, resulting in mappers not being called. Fixed in 6a31b24.

I also found more deprecation warnings - form field types should be now referenced with fully-qualified type class names, so instead of e.g. `$builder->add('identifier', 'text')` we should use `$builder->add('identifier', TextType::class)`. I'll fix that as well.

andrerom:
> I also found more deprecation warnings - form field types should be now referenced with fully-qualified type class names, so instead of e.g. $builder->add('identifier', 'text') we should use $builder->add('identifier', TextType::class). I'll fix that as well.

hmm, that seems like a bc break.
@glye any input on if this was intended change or not? or should we wait to check with @bdunogier once back?

glye:
@andrerom Yes, I think this is the intention, and I believe we have done it partially already. Symfony allows both ways since 2.7: http://api.symfony.com/2.7/Symfony/Component/Form/FormBuilder.html#method_add
but I think the full names are preferred - they are used in documentation since 2.8.
http://symfony.com/doc/2.8/reference/forms/types/text.html

alongosz:
@andrerom, @glye: I looked into this just now. There are two issues here:

1. Although it's available since `2.7`, using it with a Symfony 2.7 web app (e.g: `->add('task', TextType::class)` doesn't work for me (it throws an exception: `Could not load type "Symfony\Component\Form\Extension\Core\Type\TextType"`). Maybe it is fully supported since 2.8.

2. Notation `ClassName::class` is available since PHP `5.5`. However we require at least PHP `5.5` for eZ Platform `1.4`, so we should be safe there.

Just to be safe I'll add a requirement of PHP `5.5` to `composer.json`.

**Question**: Should I bump Symfony components to 2.8 [here](https://github.com/ezsystems/repository-forms/blob/master/composer.json#L13-L14) as well?

glye:
> Should I bump Symfony components to 2.8 here as well?

This mention says yes: https://github.com/FriendsOfSymfony/FOSUserBundle/issues/1985#issuecomment-161201861

Though it seems to contradict the docs, weird. If we can go to 2.8, that's an easy way out, anyway.

andrerom:
it's fine to bump to symfony 2.8 + in master.

lib/Form/Type/Content/ContentEditType.php
```diff
@@ -19,26 +22,26 @@
  */
 class ContentEditType extends AbstractType
 {
-    public function getName()
+    public function getBlockPrefix()
```
alongosz: This might be a BC break. 
Looking at [Symfony 2.8 upgrade notes](https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md#form) I could replace this with:
```
public function getName()
{
    return $this->getBlockPrefix();
}
```
to keep BC, but `getName` should rather be removed (according to notes mentioned above) or return fully-qualified class name [which is already done by `AbstractType`](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Form/AbstractType.php#L65-L72).
Ping @andrerom @glye

glye: Upgrade notes says: "If you want to create types that are compatible with Symfony 2.3 up to 2.8 and don't trigger deprecation errors, implement both getName() and getBlockPrefix()". So that seems to be the way to go. `getName()` should then be as in your example.

alongosz: Ok, fixed in 3732273.

andrerom:
@glye this seems to fail after https://github.com/ezsystems/ezpublish-kernel/pull/1749 was merged.

glye:
Checking... So this test isn't run for kernel PRs, apparently.

glye:
@andrerom @alongosz Found it. In repository-forms, error parameters are %-wrapped here:
https://github.com/ezsystems/repository-forms/blob/master/lib/Validator/Constraints/FieldTypeValidator.php#L43

This doesn't happen in REST, hence the inconsistency. Since the code is by JV, and he +1-ed #1749, I think the right thing is to undo the %-wrapping here. See new PR: https://github.com/ezsystems/repository-forms/pull/94

glye:
@alongosz https://github.com/ezsystems/repository-forms/pull/94 has been merged, so please rebase on master and it should work. Pardon the hassle.

alongosz:
@glye thx, it works :)

andrerom:
+1

bundle/DependencyInjection/Compiler/FieldTypeFormMapperDispatcherPass.php
```diff
@@ -54,7 +54,7 @@ private function findTaggedFormMapperServices(ContainerBuilder $container)
         $formMappers = $container->findTaggedServiceIds('ez.fieldType.formMapper');
         foreach (array_keys($formMappers) as $serviceId) {
             @trigger_error(
-                'The ez.fieldType.formMapper service tag from $serviceId is deprecated in ezsystems/repository-forms 1.3, ' .
+                "The ez.fieldType.formMapper service tag from $serviceId is deprecated in ezsystems/repository-forms 1.3, " .
```
glye: Good catch :)

glye:
+1. Also very orderly commits, @bdunogier would be proud ;)

andrerom:
Thanks @alongosz and @glye merged :)

PR #94, opened 2016-08-15
Fix EZP-26148: String length validation error contains %-symbols

glye:
> Follow up to https://jira.ez.no/browse/EZP-26148
> Status: Ready for review

During https://github.com/ezsystems/repository-forms/pull/93 it was found that https://github.com/ezsystems/ezpublish-kernel/pull/1749 introduced a regression in repoforms. It turns out that repoforms automatically wraps field error parameters in `%`s, while kernel REST doesn't. So the fix in the kernel meant we got double-wrapped params in repoforms.

This undoes the wrapping in repoforms, while simplifying the code.

**Alternative:** Introduce automatic %-wrapping in REST. I'm not a fan, because a) KISS, and b) Symfony docs assume no automagic, let's stay in sync with that: http://symfony.com/doc/2.7/components/translation/usage.html#message-placeholders

### DIFF
--- a/lib/Validator/Constraints/FieldTypeValidator.php
+++ b/lib/Validator/Constraints/FieldTypeValidator.php
@@ -39,9 +39,7 @@ abstract class FieldTypeValidator extends ConstraintValidator
             $message = $error->getTranslatableMessage();
             /** @var \Symfony\Component\Validator\Violation\ConstraintViolationBuilderInterface $violationBuilder */
             $violationBuilder = $this->context->buildViolation($message instanceof Plural ? $message->plural : $message->message);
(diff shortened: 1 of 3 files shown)


andrerom:
+1, but im +1 either way, so maybe quickly settle this with bd on monday

andrerom:
@lolautruche as you +1'ed the kernel pr, we assume you also agree with the changes here? *(the code here where originally made by you hence the question ;) )*

lolautruche:
Of course I'm +1 😉

andrerom:
Then we have our answer @glye ;)   *(squash and push issue back to QA once merged btw, I moved it back to review because of open question above)*

PR #96, opened 2016-10-12
EZP-26410: Fixed exception; Runtime Notice

nspyke:
- Resolved: Runtime notice: Only variables should be passed by reference
- Created related tests for those classes

Screenshot

### DIFF
--- a/tests/RepositoryForms/FieldType/Mapper/SelectionFormMapperTest.php
+++ b/tests/RepositoryForms/FieldType/Mapper/SelectionFormMapperTest.php
@@ -7,155 +7,43 @@
  */
 namespace EzSystems\RepositoryForms\Tests\FieldType\Mapper;
 
-use eZ\Publish\API\Repository\FieldType;
-use eZ\Publish\API\Repository\FieldTypeService;
-use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
-use EzSystems\RepositoryForms\Data\Content\FieldData;
+use eZ\Publish\Core\Repository\Values\ContentType\FieldDefinition;
 use EzSystems\RepositoryForms\FieldType\Mapper\SelectionFormMapper;
-use Symfony\Component\Form\FormConfigInterface;
-use Symfony\Component\Form\FormFactoryInterface;
-use Symfony\Component\Form\FormInterface;
 
-class SelectionFormMapperTest extends \PHPUnit_Framework_TestCase
+class SelectionFormMapperTest extends BaseMapperTest
 {
-    private $fieldTypeService;
-
-    protected function setUp()
-    {
-        $this->fieldTypeService = $this->getMockBuilder(FieldTypeService::class)
-            ->getMock();
-        $this->fieldTypeService
-            ->expects($this->any())
-            ->method('getFieldType')
-            ->willReturn($this->getMockBuilder(FieldType::class)->getMock());
-    }
-
     public function testMapFieldValueFormNoLanguageCode()
     {
         $mapper = new SelectionFormMapper($this->fieldTypeService);
 
-        $config = $this->getMockBuilder(FormConfigInterface::class)->getMock();
-        $config->expects($this->once())
-            ->method('getOption')
-            ->with('languageCode')
-            ->willReturn(false);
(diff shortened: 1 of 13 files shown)


tests/RepositoryForms/FieldType/Mapper/SelectionFormMapperTest.php
```diff
@@ -0,0 +1,154 @@
...
+        $fieldForm = $this->getMockBuilder(\Symfony\Component\Form\FormInterface::class)->getMock();
+        $fieldForm->expects($this->once())
+            ->method('getConfig')
+            ->willReturn($config);
+
+        $data = $this->getMockBuilder(\EzSystems\RepositoryForms\Data\Content\FieldData::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $fieldDefinition = $this->getMockBuilder(\eZ\Publish\API\Repository\Values\ContentType\FieldDefinition::class)
```
bdunogier: In general, I would avoid mocking value objects, and would just create instances. It would greatly shorten and simplify the unit test.

tests/RepositoryForms/FieldType/Mapper/TextBlockFormMapperTest.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        $mapper = new TextBlockFormMapper($this->fieldTypeService);
+
+        $config = $this->getMockBuilder(\Symfony\Component\Form\FormConfigInterface::class)->getMock();
+        $config->expects($this->once())
+            ->method('getOption')
+            ->with('languageCode')
+            ->willReturn(false);
+
+        $formFactory = $this->getMockBuilder(\Symfony\Component\Form\FormFactoryInterface::class)
```
bdunogier: Why not import those classes ? They're used several times in the test, after all, and I don't think it would reduce readability to do so.

nspyke:
I've updated the tests to use 'Use...' statements.
Regarding the value object mock, I've always practised that you should only ever test the SUT and mock everything else.

bdunogier:
@nspyke we don't mock value objects, as a matter of fact. It is one of _those topics_ where I guess there are strong debates, but if you search the web for "don't mock value objects", you'll see that we are not making things up :-)

It is one of the advantages of immutable, logic-less values: you can just use them. You wouldn't mock an array, would you ?

nspyke:
@bdunogier I conceed. No, I wouldn't mock an array. 
Is there an implementation of eZ\Publish\API\Repository\Values\ContentType\FieldDefinition I could use as the mocked value object is an abstract class with abstract methods?

bdunogier:
You can use `eZ\Publish\Core\Repository\Values\ContentType\FieldDefinition`.

nspyke:
@bdunogier I've done the requested updates. When can this be merged?

andrerom:
With half of Europe off until Wednesday, sometime by the end of the week in time for 1.5.2/1.6.1 rc's next week. But maybe @glye can have look as well in between :)

_Btw, seems you're not using same author email on commits as attached to your github account, on purpose?_

nspyke:
> Btw, seems you're not using same author email on commits as attached to your github account, on purpose?

No. Not on purpose. It's because this was done on my work computer, and Git is configured to use my work email address.

janit:
+1

bdunogier:
@nspyke could you please squash the commits into one ? We will if you won't :)

Meantime, I'll send this for QA. What was your case for getting this error to happen again, @nspyke ?

nspyke:
The error occurred when the user content type did not have a defined field name for a non eng-GB locale.
In my case, I was using dut-NL.
I'll leave the squashing to you. Every time I try to squash commits, I end up making it worse by making more.

miguelcleverti:
PR approved by QA

andrerom:
> I'll leave the squashing to you. Every time I try to squash commits, I end up making it worse by making more.

No problem, we have it right here in the interface now, so no worries. But it means it will end up as one commit, which for larger pull requests don't always fits either.

bdunogier:
Merged to 1.4 and master @ 3977a7976e8d78002a907d37529b396bb0aba1f1.

Thank you @nspyke.

PR #97, opened 2016-10-17
Allow Symfony 3.0

bdunogier:
> Fixes [EZP-26234](http://jira.ez.no/browse/EZP-26234)

While the kernel accepts Sf 3.0, this package doesn't, thus blocking its installation.

Opened against 1.4 as master is still on this branch.


bdunogier:
Replaced by https://github.com/ezsystems/repository-forms/pull/98.

PR #98, opened 2016-10-17
Use symfony/symfony in require section

emodric:
Reasoning:

1) This would allow to test eZ Platform 1.5 with Symfony 3.x
2) `require` section is wrong anyhow, since it's missing some other dependencies, like Twig or Symfony Event Dispatcher component, for stuff available in `lib`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,7 @@
     "require": {
         "php": "~5.5|~7.0",
         "ezsystems/ezpublish-kernel": "^6.4",
-        "symfony/form": "~2.8",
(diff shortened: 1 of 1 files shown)


emodric:
Alternatively, we can require `"symfony/symfony": "^2.7 | 3.0"`

bdunogier:
I'd rather go for a requirement of the full Symfony. The bundle *does* require them, and shouldn't depend on the kernel for that.

emodric:
@bdunogier Okay, done.

PR #100, opened 2016-11-28
JMS translation bundle integration

bdunogier:
> EZP-26704

- Custom extractor for fieldtypes names
- Forced domains in mapper classes for extraction
- Regenerated strings


bdunogier:
I'm getting the same failure locally with only repository-forms on the jms branch.

bdunogier:
One failure left due to PHP 5.5 still being tested, needs to be removed in master.

PR #101, opened 2016-11-30
EZP-26728: removed PHP 5.5 from travis matrix

bdunogier:
> [EZP-26728](http://jira.ez.no/browse/EZP-26728)


andrerom:
Should update composer.json to

bdunogier:
> Should update composer.json to

Done.

PR #108, opened 2017-01-17
EZP-26835: Remove french translations will be handle in ezplatform-i18n

nicolas-bastien:
Jira : https://jira.ez.no/browse/EZP-26835

# Description

Remove french translations will be handle in ezplatform-i18n


bdunogier:
We will have to upload those to the crowdin project, though. But we can get them from the last release.

nicolas-bastien:
@bdunogier as we had a hard time making synchronization to work, I won't be for uploading but I would rather translate in the interface dirrectly.

I'll do that.

nicolas-bastien:
done just a couple of udw missing

PR #111, opened 2017-02-20
Spike: content edit

bdunogier:
Part of [EZP-27001](https://jira.ez.no/browse/EZP-27001)

See the commit messages for more details.

Prototypes creating a draft from an existing content (`/content/create/draft/{contentId}/{fromVersionNo}/{fromLanguage}/{toLanguage}`), and editing/publishing that draft (`/content/edit/{contentId}/{versionNo}/{language}`).

### DIFF
--- a/doc/specifications/drafts/content_edit.md
+++ b/doc/specifications/drafts/content_edit.md
@@ -1,0 +1,29 @@
+# Symfony forms content edit
+
+## Synopsis
+repository-forms currently supports creating content without an intermediate draft.
+
+Support must be added for editing content. It implies creating a new version of an existing content,
+rendering a form with the previous version's fields values, and allowing the user to save the draft,
+cancel or publish the new version. It should also allow choosing a language, either for translation
+or 
+
+Ideally, the draft would only be created when we need it, not on the first display.
+However, any interactive action on Fields (custom buttons, etc) will require that the draft
+is created.  Note that the exact same problem exists in createWithoutDraft. We just don't
+have any interactive FieldTypes at the moment.
+
+## Background research
+
+### How repository forms does it for content types 
+To provide edition of ContentTypes, `\EzSystems\PlatformUIBundle\Controller\ContentTypeController::updateContentTypeAction`
+will attempt to load a draft for the cotnent, and create one if there isn't. But this only works because
+ContentTypes have a unique draft mechanism. Content drafts have a different approach, where versions
+_belong_ to a user and a user may have several drafts of the same content at a given time.
+
+### How legacy does it for content
+In legacy, an action is responsible for creating a new draft. It will by default do so
+for the currently published version, but has an option to use a different version as a source.
+If not specified, the target language and translation mode (edit or translate) will be asked.
+After creating the draft, it will redirect the user to another action where the draft is edited, then
+either cancelled, saved or published.
--- a/doc/specifications/drafts/content_edit.md
+++ b/doc/specifications/drafts/content_edit.md
@@ -62,6 +62,26 @@ Uses the same exact form than the content edit one.
 
 #### Architecture
 The Form, `ContentEditType`, uses a `ContentUpdateData` object.
+
+## Additional topics
+
+### Form rendering customization
+In order to benefit from the save level of flexibility legacy provided, the first thing that comes
+to mind is views: if we have a ContentEditView, we can easily override the template and customize
+the form using [form theming](https://symfony.com/doc/current/form/form_customization.html#form-theming-in-twig).
+
+See https://github.com/ezsystems/ezplatform-demo/blob/master/app/Resources/views/user/registration_content_form.html.twig
+for an example of a content form customization.
+
+This involves:
+- ViewBuilders
+- ViewProviders + configuration
+- Default templates
+- Practical examples (anything in `ez*-demo` ?)
+
+> Very down-to-earth idea: if `fieldsData` was indexed by field def identifier, it would simplify
+> the template code a lot.
+
 ## Background research
 
 ### How repository forms does it for content types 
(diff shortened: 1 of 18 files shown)


doc/specifications/drafts/content_edit.md
```diff
@@ -0,0 +1,29 @@
+# Symfony forms content edit
+
+## Synopsis
+repository-forms currently supports creating content without an intermediate draft.
+
+Support must be added for editing content. It implies creating a new version of an existing content,
+rendering a form with the previous version's fields values, and allowing the user to save the draft,
```
andrerom: nitpick: when you create a draft it typically contain the values from previous version initially, so the form won't have to care for this itself, it should just edit the draft all the way until user hits publish.

doc/specifications/drafts/content_edit.md
```diff
@@ -0,0 +1,29 @@
...
+cancel or publish the new version. It should also allow choosing a language, either for translation
+or 
+
+Ideally, the draft would only be created when we need it, not on the first display.
```
andrerom: >  the draft would only be created when we need it, not on the first display.

depends, this can break the UI protection we have against several editing draft.

So imho what we should probably aim for is that API takes care about this, creating internal draft when you can for creating draft, and not change it to real draft before updates on it happens _(on store)_.

doc/specifications/drafts/content_edit.md
kamilmusial: > so the form won't have to care for this itself

But the form fields must be filled already anyway. Otherwise `save draft` would pass empty data, wouldn't it?

doc/specifications/drafts/content_edit.md
bdunogier: > So imho what we should probably aim for is that API takes care about this, creating internal draft when you call for creating draft, and not change it to real draft before updates on it happens (on store).

Okay. Can I consider that it will work like this for this prototype, or should I put this on hold ?

doc/specifications/drafts/content_edit.md
bdunogier: > it should just edit (update) the draft when user hits save draft.

This is exactly what it does. I'll rework this document, it is misleading.

doc/specifications/drafts/content_edit.md
andrerom: > What ?

As in in legacy, and afaik in Platofmr UI, a draft is created so that if someone else hits edit button on same content edit, the UI can show a conflict warning.

bdunogier: The warning shows up when the _user_ has a draft, does it not ? It will tell you if you already have a draft of this content, and will ask if you want to edit it, or create a new one. As far as I know, you're not warned if somebody else has a draft.

Note that this part isn't even mentioned in the requirements now. It is indeed an important part, and I'm trying to keep it in mind. Version/conflict management was only added to PlatformUI a couple weeks ago.

andrerom: ok, lets make sure it is clarified there with PM/BF then.

papelipe: I would define the requirements like:
- Initial version of the object (v0) should not require a draft to be created before store.
- Edits of existing versions should create a draft as it is needed for checking if it is already being worked on etc. as AR mentions.

The overal idée is to not create a draft until you actually publish something. This is mainly to remove a potential DOS attack and also create many "loose drafts" whenever users are e.g. registering. Can easily be exploited to fill up the DB.

bdunogier: > This is mainly to remove a potential DOS attack and also create many "loose drafts" whenever users are e.g. registering. Can easily be exploited to fill up the DB.

Creating a new (v0) content without creating a draft is already possible.
We had decided to shape it that way to limit the risk of the DB being filled up.

> Edits of existing versions should create a draft as it is needed for checking if it is already being worked on etc. as AR mentions.
> The overal idée is to not create a draft until you actually publish something.

These two sentences are contradicting each other, I'm afraid.
We can create the draft when a content edit (version > 1) is displayed, or try to do it later.
For now, the Public API and REST API don't do anything in this regard. PlatformUI has its own logic for it, a bit like what Legacy was doing, but afaik it doesn't prevent a user from creating multiple drafts of the same content.

Is version / conflict management something we want to integrate into the current spikes or not ?

bdunogier:
Added MapLocation FieldValue form implementation, as a simple approach to complex forms.

doc/specifications/drafts/content_edit.md
```diff
@@ -0,0 +1,102 @@
...
+
+- path: `/content/create/draft/{contentId}/{fromVersionNo}/{fromLanguage}/{toLanguage}`
+- controller action: `ContentEditController::createContentDraftAction`
+
+All the arguments except for `contentId` are optional. Arguments that are specified will
+be used as default values for the form. The draft will only be created when the form is 
+submitted. However, this resource is meant to be used by other parts of the system to create 
+a content draft.
+
+When the form is submitted and a draft is created, a redirection to the content editing route
```
glye: Even when publishing directly? A draft should be created in any case, but since the edit draft route is described as showing the edit form: If you publish right after creating a draft (optionally changing some fields), and validation passes, there is no need to show the form again (whether that means skipping the edit draft route, or skipping its form view, is tbd).

doc/specifications/drafts/content_edit.md
```diff
@@ -0,0 +1,102 @@
...
+Will load the requested content, in the requested version, and show the edit form for the fields.
+Uses the same exact form than the content edit one.
+
+#### Architecture
+The Form, `ContentEditType`, uses a `ContentUpdateData` object.
+
+## Additional topics
+
+### Form rendering customization
+In order to benefit from the save level of flexibility legacy provided, the first thing that comes
```
glye: "same level"

sunpietro:
What's the status of this PR?

andrerom:
It was afaik the prototype for content edit feature that was added later, so this can be closed. If something is missing from this it can be extracted, cleaned up, and opened on a new PR.

PR #112, opened 2017-02-20
EZP-25007: "Number of text rows in the editor" should be removed from RichText fieldtype

alongosz:
> Fixes [EZP-25007](https://jira.ez.no/browse/EZP-25007)
> **IMPORTANT**: It has to be merged along with [Kernel PR #1915](https://github.com/ezsystems/ezpublish-kernel/pull/1915).

This PR removes `numRows` setting from the `RichText` Field Type and will also have its ezpublish-kernel counterpart.

**TODO**:
- [x] Remove all references of `numRows` setting from repository-forms.
- [x] Create corresponding ezpublish-kernel PR ([#1915](https://github.com/ezsystems/ezpublish-kernel/pull/1915)).
- [ ] Create test (?).

### DIFF
--- a/bundle/Resources/translations/ezrepoforms_content_type.en.xlf
+++ b/bundle/Resources/translations/ezrepoforms_content_type.en.xlf
@@ -414,12 +414,6 @@
         <note>key: field_definition.ezpage.default_layout</note>
         <jms:reference-file line="45">lib/FieldType/Mapper/PageFormMapper.php</jms:reference-file>
       </trans-unit>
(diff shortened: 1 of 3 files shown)


alongosz:
The remaining question here is if I should create BDD for this. It seems to me that there's no coverage for editing of any Field Definition for Content Type. Field Types are covered only with regard to setting value for Content Object.

Also worth pointing out - even if I create BDD it could be difficult to make it passing due to kernel dependency (as mentioned [here](https://github.com/ezsystems/ezpublish-kernel/pull/1915#issuecomment-281324961)). I'm up for a challenge though :)

andrerom:
BDD question can be handled separately from this PR, it's a topic of it's own.

PR #113, opened 2017-02-21
JSON schema report

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-27005

This report provides which information are important to render field types from frontend perspective.


andrerom:
_Side: [JSON API is described in JSON Schema](http://jsonapi.org/faq/#is-there-a-json-schema-describing-json-api), but to be noted as far as I can see it is considered [an addition to RAML](https://github.com/json-api/json-api/issues/271) specifically for use by javascript based API consumer clients. _

PR #114, opened 2017-02-25
Prototyped JSONSchema content edit forms

bdunogier:
Go to http://site/content/create/nodraft/folder/eng-GB/2.json. It will respond with the jsonschema
to create a new folder:

```json
{
  "type": "object",
  "properties": {
    "fieldsData": {
      "type": "array",
      "items": [
        {
          "title": "title",
          "type": "object",
          "properties": {
            "value": {
              "type": "string",
              "title": "Title",
              "propertyOrder": 1
            }
          },
          "required": [
            "value"
          ]
        },
        {
          "title": "short_title",
          "type": "object",
          "properties": {
            "value": {
              "type": "string",
              "title": "Short title",
              "propertyOrder": 1
            }
          }
        },
        {
          "title": "author",
          "type": "object",
          "properties": []
        },
        {
          "title": "intro",
          "type": "object",
          "properties": []
        },
        {
          "title": "body",
          "type": "object",
          "properties": []
        },
        {
          "title": "enable_comments",
          "type": "object",
          "properties": {
            "value": {
              "type": "boolean",
              "title": "Enable comments",
              "propertyOrder": 1
            }
          }
        },
        {
          "title": "image",
          "type": "object",
          "properties": []
        }
      ],
      "propertyOrder": 1
    }
  },
  "required": [
    "fieldsData"
  ]
}
```

It uses [limenius/liform-bundle](https://packagist.org/packages/limenius/liform-bundle), that provides serialization of Symfony Forms into jsonschema.

The output has been tested with https://mozilla-services.github.io/react-jsonschema-form/, and does output a sensible form:

capture d ecran 2017-02-25 a 01 36 01

#### The very complicated form
The form that is generated is very nested: `properties > fieldsData > items > item[] > properties > value > {actual field value}`.

The resulting form, even if it has all the elements, has multiple depth level, requiring either to handle or simplify that structure. This isn't an ideal developer experience. Options to simplify:

##### Option A: simplify the structure generated by the serializer
Before passing the structure returned by LiForm to the JSON serializer, walk through it to produce a simpler version.

> Pro: shorter
> Con: ugly

##### Option B: generate a simpler form object
Reduce the nesting level in `ContentEditType` (new `FormType`). The challenge is that this nested structure is used to "map" the underlying objects to Symfony Forms. Building the Form _and_ processing POSTed data is done by mapping an object tree. Each property of the passed object is mapped to its own FormType.

The `Content` is not enough to build the form. It needs to be mapped with the `ContentType` it is an instance of to get the type of each File. This is what the `ContentEditFormMapper` does:
- it builds a `ContentUpdateData` object with the `Content` and the `ContentType`
- it iterates over the `ContentType`'s FieldDefinitions, and adds `FieldData` objects to the `ContentUpdateData`. A `FieldData` object is an aggregate of `Field`, `FieldDefinition` and th Field's `Value`.

An option would be to use an object that has Field objects, representing each edited Field, at the root level (mapped to "properties" in the form). Each `Field` object _will_ have to be mapped to a `FormType`, so that the editing UI of the FieldType can be rendered. This is where FieldMappers are used. 

> Pro: clean
> Con: complicated


nacmartin:
Hi

**Option A** has an additional problem: that the data that you will produce when the user fills the form, and that will be submitted, won't match what the Sf Form expects. Then you will need to transform again the data into the format that can be understood by the API. This is quite annoying.

**Option B** requires a deep understanding of ezplatform, that I unfortunately don't have.

There is also an **Option C**, that is to do the work on the React form generator:

In your intermediate objects you can give some kind of UI hints of how they should be generated. Like specifying that `fieldsData` should be rendered with a widget without a strong visual representation with a title.

To do so you can add properties to your JSON schema that are not part of the standard, and tools dealing with properties that are unknown should ignore them  (A JSON Schema MAY contain properties which are not schema keywords. Unknown keywords SHOULD be ignored.). In the case of LiformBundle and liform-react, I chose for instance to let the user specify a `widget` property that defines the type of widget being used. So I would add a `widget:objectHolder` property to `fieldsData`. Then, in the generator I would write a widget that just renders the children of `fieldData` without displaying the label.

Mozilla's generator does this with an additional UISchema that gives UI hints. I don't like this approach very much because it means to write _another_ serializer, but it is possible way. It is a matter of taste.

Anyways, whether the UI hints are part of the schema or are supplied as an additional JSON object, you can pretty much avoid representing intermediate object holders, or customising its appearance so they are just <divs> with different styles to represent sections or something like that. It can be via widget definition, or via providing a property `showLabel:false` or something like that.

In any case, unless you are 100% happy with the themes and widgets that generators,  at some point there is a need for writing widgets and customising the generator somehow.

Btw, if you end up having very special needs, to write your own react form generator is not something crazy: react-liform consists in just 5 small files (if you don't count the theme/widgets) https://github.com/Limenius/liform-react/tree/master/src

bdunogier:
Thank you for your detailed answer, @nacmartin.

I have actually given option B a try, and it turned out alright so far. I might have forgotten something, as the edit use-case has a few key differences with the create one, but I'll test that asap. In the meanwhile, the very simplified form serializes like a charm.

Customizing rendering is indeed an option, but it has a small defect: I've been toying around with matthias noback's console-form, and it also has issues because of this nested form. This indicates, I think, that most form related libraries will have issues with that. And I don't like issues :)

The latest commit contains my attempt.

adamwojs:

PR #116, opened 2017-04-06
Fix EZP-27221: it is not possible to empty the urlAliasSchema field

bdunogier:
> Fixes [EZP-27221](https://jira.ez.no/browse/EZP-27221)

The [Repository's mapper](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/Repository/Helper/ContentTypeDomainMapper.php#L145-L145) will ignore the value if it is set to null. This sets the `empty_value` for the urlAliasSchema to false so that it is not ignored.


bdunogier:
Ping @ramzi-arfaoui.

dpobel (approved):
(you can rebase the branch to get rid of the problem reported by @ezrobot)

bdunogier:
Thank you @dpobel.

yannickroger (approved):

PR #117, opened 2017-04-07
Fix EZP-25923: Creating a content type without attributes raise a 503 Service Unavailable

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25923

# Description

This patch disables the publish button (*OK*) until the Content Type has at least a field definition. While at it, I also disable the *Remove selected field definitions* button under the same condition.

# Test

manual tests


bdunogier:
Given that the change modifies the behaviour of buttons in general, I'd simplify it a little bit. What about `Fix EZP-25923: disable content-type form buttons when not usable` ?

PR #119, opened 2017-04-29
EZP-26878: Add "Default value" input field for Checkbox, Float, ISBN,…

adamwojs:
… and Integer content definitions in EDIT mode

> JIRA: https://jira.ez.no/browse/EZP-26878

# Description 
This PR adds missing "Default value" field in definition of following content fields:
- Checkbox
- Float
- Integer
- ISBN **

** _I'm not sure there is any real life use case for default value in ISBN content field_


lib/FieldType/DataTransformer/CheckboxValueTransformer.php
```diff
@@ -0,0 +1,35 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryForms\FieldType\DataTransformer;
+
+use eZ\Publish\Core\FieldType\Checkbox\Value;
+use Symfony\Component\Form\DataTransformerInterface;
+
+class CheckboxValueTransformer implements DataTransformerInterface
```
bdunogier: I'm thinking that those transformers (Checkbox, ISBN, Integer, Float) could be replaced with a unique `SimpleValueTransformer` (not the best name). It would look like this:

```php
<?php
use Symfony\Component\Form\DataTransformerInterface;

class SimpleValueTransformer implements DataTransformerInterface
{
    private $valueClass;
    
    public function __construct($valueClass)
    {
        $this->valueClass = $valueClass;
    }

    public function transform($value)
    {
        if (!$value instanceof $this->valueClass) {
            return null;
        }

        return $value->value;
    }
    public function reverseTransform($value)
    {
        if (!$value) {
            return null;
        }
        
        return new $this->valueClass($value);
    }
}
```

A mapper would use it like this:
```php
->addModelTransformer(new SimpleValueTransformer(CheckboxValue::class))
```

It would allow us to replace four classes with one that covers many simple cases.

What do you think ?

adamwojs: @bdunogier I'm agree. I think we can even replaces those transformers with EzSystems\RepositoryForms\FieldType\DataTransformer\FieldValueTransformer, behind the scene they are doing that same thing (I've checked and it works fine).

I assume this changes should be pushed in two separate commits: 1st  with implementation for checkbox, integer, float and isbn content field types, and 2nd with replacment of TextLineValueTransformer  in favor FieldValueTransformer in TextLineFormMapper. Am I right ?

andrerom: @adamwojs yes separate commits are good, we can always squash on merge if we don't need the history in master itself.

bdunogier: > I assume this changes should be pushed in two separate commits: 1st with implementation for checkbox, integer, float and isbn content field types, and 2nd with replacment of TextLineValueTransformer in favor FieldValueTransformer in TextLineFormMapper. Am I right ?

Aye. I mean yes :)

adamwojs:
@bdunogier, @andrerom I updated the PR according to suggestions from CR ;-)

bundle/Resources/translations/ezrepoforms_content_type.en.xlf
```diff
@@ -276,6 +282,12 @@
         <note>key: field_definition.ezdatetime.use_seconds</note>
         <jms:reference-file line="28">lib/FieldType/Mapper/DateTimeFormMapper.php</jms:reference-file>
       </trans-unit>
+      <trans-unit id="ed0d3cb0a19dd96c1708957f17e15204465c69b6" resname="field_definition.ezfloat.default_value">
```
andrerom: @bdunogier we don't reuse string for these content type fieldDefinition global attributes?

PR #120, opened 2017-05-10
Fix EZP-25923: disable content-type form buttons when not usable

tomaszmadeyski:
# Description
This is a port of https://github.com/ezsystems/repository-forms/pull/117 for `1.5` version
This fixes the https://jira.ez.no/browse/EZP-25923 for `Platform 1.7` as this version relies on `Repository-Forms` in version `1.5`

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,27 +1,31 @@
 <?php
 
-return Symfony\CS\Config\Config::create()
-    ->setUsingLinter(false)
-    ->setUsingCache(true)
-    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
-    ->fixers([
-        'concat_with_spaces',
-        '-concat_without_spaces',
-        '-empty_return',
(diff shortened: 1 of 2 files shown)


andrerom:
Seems you need to port php-cs-fixer config update to.

tomaszmadeyski:
@andrerom done in f1084d3 - now it discovers some errors but this is not a stuff I changed, it was there before - it is not a subject of this PR. Question is should I fix it in this PR or just leave it as it was before

alongosz:
I would rather suggest moving fixer update to a separate PR along with fixes for these discovered errors. Then this PR could be just rebased.

andrerom:
For referance with the cs fixes I was +1

@alongosz was the changes ok with you besides the CS topic?
ref: 92c2517be1faf3b4ccf6c6e0deaeea7bba368195

Ping me if not and I'll revert straight away.

PR #122, opened 2017-05-10
EZP-25197: Missing validation of the role form

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-25197

# Description

This PR adds missing validation of the role form, especially:

* Uniqueness of the role name (fixed incorrect FQN of RoleData class in validation.yml)
* Not blank role name (currently there is only required attribute on `<input />` tag)
* Max length of the role name

### DIFF
--- a/bundle/Resources/translations/validators.en.xlf
+++ b/bundle/Resources/translations/validators.en.xlf
@@ -36,6 +36,11 @@
         <target>Section identifier may only contain letters from "a" to "z", numbers and underscores.</target>
         <note>key: ez.section.identifier.format</note>
       </trans-unit>
+      <trans-unit id="963a415cd08151ecf8726e651a7a36140e05f4b5" resname="ez.role.identifier.unique">
+        <source>The role identifier "%identifier%" is used by another role. You must enter a unique identifier.</source>
+        <target>The role identifier "%identifier%" is used by another role. You must enter a unique identifier.</target>
+        <note>key: ez.role.identifier.unique</note>
+      </trans-unit>
     </body>
   </file>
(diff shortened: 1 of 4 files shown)


bundle/Resources/translations/validators.en.xlf
```diff
@@ -36,6 +36,11 @@
         <target>Section identifier may only contain letters from "a" to "z", numbers and underscores.</target>
         <note>key: ez.section.identifier.format</note>
       </trans-unit>
+      <trans-unit id="963a415cd08151ecf8726e651a7a36140e05f4b5" resname="ez.role.identifier.unique">
+        <source>Non unique role identifier "%identifier%"</source>
```
glye: Not a sentence. I'd explain it more clearly, with something like: "The role identifier 'blarghhh' is used by another role. You must enter a unique identifier."

adamwojs:
PR updated with correction of the grammatical errors. Definitely grammar of natural languages is not my strong point ;-)

adamwojs:
PR updated with more clearly error explanation suggested by @glye. 

@glye Yes, because `NotBlank` and `Length` constraints are part of Symfony Validator Component. Tests for forms themselves are in platform-ui repository.

andrerom:
@glye ok with the updated string? :)

andrerom:
@adamwojs  Did you check if the behat scenario refered to in issue can/must be improved as well now?

adamwojs:
@andrerom I didn't, but almost for sure, it needs to be enabled and there is a missing scenario for max length constraint. I can take care of this today evening ;-)

PR #124, opened 2017-05-16
EZP-26608 adding translation for "Metadata"

tomaszmadeyski:
This a follow up on https://github.com/ezsystems/ezpublish-kernel/pull/1988 

Adding translation for "Metadata" which was a missing item on default Field Group list

Related Jira issue: https://jira.ez.no/browse/EZP-26608

### DIFF
--- a/bundle/Resources/translations/ezplatform_fields_groups.en.xlf
+++ b/bundle/Resources/translations/ezplatform_fields_groups.en.xlf
@@ -6,6 +6,11 @@
                 <source>content</source>
                 <target>Content</target>
             </trans-unit>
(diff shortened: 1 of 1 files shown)


bdunogier:
Is this meant to be a bugfix ? If yes, we may wanna merge it to 1.5 (the version that shipped with eZ Platform 1.7 LTS).

tomaszmadeyski:
@bdunogier I rebased it to branch 1.5. Errors which are here reported by `phpcsfixer` are because of wrong cs-fixer configuration and it will be fixed after #121 merge

PR #127, opened 2017-06-18
EZP-24800 Support allowed classes limitation for object relation fields

slaci:
This is related to this PR: https://github.com/ezsystems/ezpublish-kernel/pull/2022

I made a base class for the relation forms instead of duplication. I hope this is not a problem.

It adds a new translation... dunno whats the proper way for that. I updated the file by hand.

### DIFF
--- a/lib/FieldType/Mapper/AbstractRelationFormMapper.php
+++ b/lib/FieldType/Mapper/AbstractRelationFormMapper.php
@@ -1,0 +1,53 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\FieldType\Mapper;
+
+use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\Core\Helper\TranslationHelper;
+use EzSystems\RepositoryForms\FieldType\FieldDefinitionFormMapperInterface;
+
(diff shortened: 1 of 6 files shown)


bdunogier:
Thank you for this, i'll have a look tomorrow !

As for translations, you can run bin/extract-translations.sh from the repository-forms directory.

andrerom:
+1, but 'I'll let the reviewers give the proper review :)

slaci:
> As for translations, you can run bin/extract-translations.sh from the repository-forms directory.
Thanks for the tip. I ran this command and it updated some line references and moved my new entry to an other line... so I let it happen. I had to move the `configureOptions` method into both children classes, cause the extractor does not handle parent classes (at least for this), so it created a messages.en.xlf without this.

slaci:
I created a separate commit for the translation file. Nothing else changed (I hope 🙈 )

lib/FieldType/Mapper/AbstractRelationFormMapper.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $this->contentTypeService = $contentTypeService;
+        $this->translationHelper = $translationHelper;
+    }
+
+    /**
+     * Fill a hash with all content types and their ids.
+     * @return array
+     */
+    protected function getContentTypeHash()
```
bdunogier: Plural, `getContentTypesHash()`.

andrerom: @slaci can you fix this nitpic? :)

andrerom:
I'll handle nitpick as followup commit, thanks @slaci.

PR #128, opened 2017-06-20
Enabled PHPUnit 5

MarioBlazek:


andrerom:
Merging as this is about `require-dev` so won't affect packages using this.

MarioBlazek:
@bdunogier we do not require v5 as such. v5 is going forward with ezpublish-kernel, it drops support for PHP 5.4 and 5.5, only 5.6 and newer are supported.

PR #130, opened 2017-06-21
Added base README with badges

MarioBlazek:


README.md
```diff
@@ -0,0 +1,6 @@
+# Repository forms
+
+[![Build Status](https://img.shields.io/travis/ezsystems/repository-forms.svg?style=flat-square)](https://travis-ci.org/ezsystems/repository-forms)
+[![Downloads](https://img.shields.io/packagist/dt/ezsystems/repository-forms.svg?style=flat-square)](https://packagist.org/packages/ezsystems/repository-forms)
+[![Latest stable](https://img.shields.io/packagist/v/ezsystems/repository-forms.svg?style=flat-square)](https://packagist.org/packages/ezsystems/repository-forms)
```
bdunogier: Weird that this one shows the rc1 and not the stable. Any clue ?
Maybe we could show the latest dev version in addition ?

alongosz: I noticed that also in case of similar Solr PR. It seems to me that this is rather a https://shields.io bug, links with `/v` should point to Release, not Pre-Release (according to doc on their webpage). Not sure if we can do something about it, but @MarioBlazek - some research welcomed ;)

MarioBlazek: Yes, `/v` is for all versions that are not marked as pre release. For pre releseases we can use `/vpre`.

MarioBlazek: This cannot detect difference in rc and standard version.

MarioBlazek: I can rename `Latest stable` to `Latest version`, WDYT @bdunogier @alongosz

andrerom: @MarioBlazek Lets do that to avoid the confusion until they fix it, as far as I can see we have properly labeled non stable versions as such.

MarioBlazek:
@alongosz @bdunogier https://github.com/badges/shields/issues/885 it look like that shields.io requires capitalized letters in versions :)

PR #132, opened 2017-07-05
[WIP] EZP-27492 Content draft edit

glye:
> Implement https://jira.ez.no/browse/EZP-27492
> Work in progress
> Based on https://github.com/ezsystems/repository-forms/pull/111

Preface:
- [x] speling fixes
- [x] fix merge conflict
- [x] fix CS issues
- [x] make tests pass

And then...:
- [x] enable existing save draft / cancel actions, verify that they work
- [ ] take out commits not strictly in scope (field type specific)
- [ ] add edit button to content view
- [ ] add edit button to content version list, for specific version


glye:
The test failure is in Behat `ContentEdit::iAmShownTheContentCreationForm()` and `thereIsARelevantErrorMessageLinkedToTheInvalidField()` which is using a minimum-length constraint. 

When testing this manually the browser won't let me submit. When removing the `minlength` attribute using the browser inspector, I can submit and get a `eZ\Publish\Core\Base\Exceptions\ContentFieldValidationException: Content fields did not validate` which makes sense.

I suppose it should catch this and reload the form, right @bdunogier?

bdunogier:
> I suppose it should catch this and reload the form, right @bdunogier?

Ideally, it would be part of the validation process, and be caught by `if ($form->isValid())` in the Controller action. The form would then not be processed, and would be displayed again.

bdunogier:
> When testing this manually the browser won't let me submit. When removing the minlength attribute using the browser inspector, I can submit and get a ContentFieldValidationException (which makes sense).

It actually fails on

```
Then I am shown the content creation form
# EzSystems\RepositoryForms\Features\Context\ContentEdit::iAmShownTheContentCreationForm()
```

The next one isn't even tried. I have checked the generated DOM, and I do have this:

```html
<div class="ezfield-type-ezstring ezfield-identifier-constrained_field">
            <fieldset>
<legend><label class="required">Constrained field</label></legend>
                    <input type="text" id="ezrepoforms_content_edit_fieldsData_constrained_field_value" name="ezrepoforms_content_edit[fieldsData][constrained_field][value]" minlength="5" maxlength="10">
</fieldset>
</div>
```

It seems that the Xpath query that we use (`div.ezfield-identifier-constrained_field input[type=text]`) doesn't match this (with this branch). As far as I know, the space separating the two indicates that any level of descendants should match.

bdunogier:
I have tried it locally, and the scenario fails for a good reason: validation doesn't work as expected. To my knowledge, it has passed in the past. I don't know if something has changed.

The expected outcome is that the error message is integrated in the form, next to each field.

glye:
Weird! When publishing a folder, `FieldValueValidator::validate()` gets called by the form `handleRequest()` here https://github.com/ezsystems/repository-forms/blob/57b52775ea51886bce05a05a5ebb1ffd3b9e2bf5/bundle/Controller/ContentEditController.php#L82

...but it accepts the value of the name field even though it's invalid according to field definition. Later, here https://github.com/ezsystems/repository-forms/blob/57b52775ea51886bce05a05a5ebb1ffd3b9e2bf5/bundle/Controller/ContentEditController.php#L85

the `TextLine\Type::validate()` gets called, and it correctly detects the invalid field, but by then it's too late, as the form has been deemed valid by the controller.

glye:
Looks like field type validateValue() was being called only when the value is empty - should be vice versa. https://github.com/ezsystems/repository-forms/pull/132/commits/d6965977c73a866d96a1dc04f45842af8615b37e

glye:
@bdunogier What do you think about scope for this? Am suspecting that translation support requires more work than a mere bugfix. How about making the "to language" field disabled, wrap up this one, and deal with translation separately?

bdunogier:
> Am suspecting that translation support requires more work than a mere bugfix.

I'd separate this for now. We need to work on the translation widget for FieldTypes, in any case.

bdunogier:
P.S. This will probably require at least a bit of rebasing, I'm afraid.

What really matters is that identified bits (FieldTypes mostly) are separated. The rest of it can be a big bunch, as long as we can describe it in ~70 characters ⛩

glye:
> The rest of it can be a big bunch, as long as we can describe it in ~70 characters

Easy. My math skills are not great, but I'm pretty sure there's less than 70 characters in "big bunch".

glye:
Continued in https://github.com/ezsystems/repository-forms/pull/133

PR #133, opened 2017-07-11
EZP-27492 Content draft edit

glye:
> Implement https://jira.ez.no/browse/EZP-27492
> Ready for review
> Based on https://github.com/ezsystems/repository-forms/pull/111 and https://github.com/ezsystems/repository-forms/pull/132

Preface:
- [x] speling fixes
- [x] fix merge conflict
- [x] fix CS issues
- [x] make tests pass

And then...
- [x] enable existing save draft / cancel actions, verify that they work
- [x] take out commits not strictly in scope (field type specific)
- [x] ~add edit button to content view~ Dealt with in other issues
- [x] ~add edit button to content version list, for specific version~ Dealt with in other issues

### DIFF
--- a/doc/specifications/content_edit.md
+++ b/doc/specifications/content_edit.md
@@ -1,0 +1,58 @@
+# Symfony forms content edit
+
+## Synopsis
+repository-forms supports creating content with and without an intermediate draft. This describes
+editing with drafts.
+
+Editing content implies creating a new version of an existing content, rendering a form with the
+previous version's fields values, and allowing the user to save the draft, cancel or publish the
+new version.
+
+
+### Creating a draft from an existing version
+
+> A route that displays a form allowing to create a draft from an existing content item.
+
+- path: `/content/create/draft/{contentId}/{fromVersionNo}/{fromLanguage}/{toLanguage}`
+- controller action: `ContentEditController::createContentDraftAction`
+
+All the arguments except for `contentId` are optional. Arguments that are specified will
+be used as default values for the form. The draft will only be created when the form is 
+submitted. However, this resource is meant to be used by other parts of the system to create 
+a content draft.
+
+When the form is submitted and a draft is created, a redirection to the content editing route
+will be issued.
+
+#### Architecture
+`ContentFormProcessor::processCreateDraft()` uses a `ContentCreateDraftType` Form.
+It manipulates a `ContentCreateDraftData` object that has the parameters to create the draft.
+The actual creation of the draft is delegated to the FormProcessor API, by firing a
+`RepositoryFormEvents::CONTENT_CREATE_DRAFT`.
+
+
+### Editing a draft
+> A route that displays a content editing form, with options to discard, save or publish.
+
+- path: `/content/edit/{contentId}/{versionNo}/{language}`
+- controller action: `ContentEditController::editContentDraftAction`
+
+`language` is optional.
+
+Will load the requested content, in the requested version, and show the edit form for the fields.
+Uses the same exact form as the content edit one.
+
+#### Architecture
+The Form, `ContentEditType`, uses a `ContentUpdateData` object.
+
+
+## Future development
+
+### Version conflits
+[Version conflicts](https://jira.ez.no/browse/EZP-25465) must be handled. There are discussions
+about moving this logic down to the API. On the other hand, it would be fairly simple to implement
+that logic in this controller.
+
+### Language handling
+Translation is currently not implemented. New drafts are created in the same language as their
+predecessors.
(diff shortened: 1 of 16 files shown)


glye:
@bdunogier Tests failing now because PHPunit 6.2 is installed for PHP7, rather than 5.7 as before. Known issue? Should be fixed in composer.json?

bdunogier:
@glye if I am not mistaken, the package has been patched recently to support it. Can you have a look at the recent history, and see if there are relevant changes you should adapt to ?

glye:
Huh, I already checked in master, but let's see..

bdunogier:
Actually, you're right, Gunnstein. I also don't really see why it would install PhpUnit 6.x given the `composer.json` contents.

glye:
You're right, there was such a recent change, but it didn't cover the `features`. I'll fix.

glye:
The `features` should be phpunit 6.x compatible now. Similar change for other tests was done in the older https://github.com/ezsystems/repository-forms/commit/0290f1884924f1a500e73b1947251fa8c9a50308

glye:
And we're green again 💚 
I'll reduce the number of commits.

glye:
@bdunogier It's down to 5 commits, and ready for review if tests pass again. (Might also squash the 3rd commit...?)

lib/Data/Content/CreateContentDraftData.php
```diff
@@ -0,0 +1,19 @@
...
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\Data\Content;
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+
+class CreateContentDraftData extends ValueObject
+{
+    public $contentId;
```
bdunogier: In retrospect, this isn't very value objectish. In our ValueObjects, properties are meant to be protected, and set using the constructor (read only VO).

Could you update the properties for that, and adapt the action ?

glye:
Squashed all except the PHPunit 6.x fix.

lib/Data/Content/CreateContentDraftData.php
glye: @bdunogier Symfony `PropertyPathMapper::mapFormsToData()` doesn't like that, it needs to write to the data.
http://api.symfony.com/3.1/Symfony/Component/Form/Extension/Core/DataMapper/PropertyPathMapper.html#method_mapFormsToData
Could workaround that with setter methods, but does it make sense?

glye:
Squashed all except the PHPunit 6.x fix, and the field value validator change.

lib/Data/Content/CreateContentDraftData.php
glye: @bdunogier We could maybe also "wrap" a readonly VO in a writeable VO for the purpose of the mapper, but again, does this make sense to do?

glye:
@bdunogier I pushed this fixup just to illustrate the problem. This gathers the parameters, uses them in the `CreateContentDraftData` constructor, and uses the value object with the form, the way I interpreted your suggestion. It fails with a `PropertyReadOnlyException` due to `PropertyPathMapper::mapFormsToData()` wanting to edit properties.

Perhaps it can be solved by reimplementing the mapper, or something like that. I haven't had time to look much into it yet.

lib/Data/Content/CreateContentDraftData.php
bdunogier: Please disregard this. It is valid for values that are set in the repository. This is input data, meant to be manipulated by the client (the mapper here).

glye:
Reverted readonly valueobject commit after discussion with @bdunogier.

doc/specifications/drafts/content_edit.md
```diff
@@ -0,0 +1,102 @@
+# Symfony forms content edit
```
bdunogier: I wouldn't merge this document as is.

Could you try to update it / write a shorter one that covers what is *actually* in *this* PR ?

glye:
@bdunogier Update: Trimmed down doc, and squashed. New doc in [doc/specifications/content_edit.md](https://github.com/ezsystems/repository-forms/pull/133/files#diff-2f0483f64cc589f00c1366085d8f17c6)

doc/specifications/content_edit.md
```diff
@@ -0,0 +1,58 @@
...
+editing with drafts.
+
+Editing content implies creating a new version of an existing content, rendering a form with the
+previous version's fields values, and allowing the user to save the draft, cancel or publish the
+new version.
+
+
+### Creating a draft from an existing version
+
+> A route that displays a form allowing to create a draft from an existing content item.
```
bdunogier: Missing word ?

glye: You mean "...allowing *you* to..."? It's common and works fine either way I think. I wouldn't change it unless a native speaker says it's borked :)

PR #134, opened 2017-07-13
EZP-26700: Implement editing support for more FieldTypes

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-26700

# Description
* `ezstring` [EZP-27627: Implement editing support for TextLine FieldType](https://jira.ez.no/browse/EZP-27627) 
* `eztext` [EZP-27644: Implement editing support for TextBlock FieldType](https://jira.ez.no/browse/EZP-27644) 
* `ezurl` [EZP-27645: Implement editing support for Url FieldType](https://jira.ez.no/browse/EZP-27645) 
* `ezemail` [EZP-27619: Implement editing support for EmailAddress FieldType](https://jira.ez.no/browse/EZP-27619) 
* `ezdatetime` [EZP-27634: Implement editing support for DateAndTime FieldType](https://jira.ez.no/browse/EZP-27634) 
* `ezdate` [EZP-27635: Implement editing support for Date FieldType](https://jira.ez.no/browse/EZP-27635) 
* `eztime` [EZP-27636: Implement editing support for Time FieldType](https://jira.ez.no/browse/EZP-27636) 
* `ezfloat` [EZP-27637: Implement editing support for Float FieldType](https://jira.ez.no/browse/EZP-27637) 
* `ezinteger` [EZP-27620: Implement editing support for Integer FieldType](https://jira.ez.no/browse/EZP-27620) 
* `ezboolean` [EZP-27632: Implement editing support for Checkbox FieldType](https://jira.ez.no/browse/EZP-27632) 

Fields were implemented using built in Symfony form types. They are editable without any JS enhancement needed.

**Support for more advanced FieldTypes will come in form of separate PRs.**

# TODO
- [x] implement views from [EZP-27578](https://jira.ez.no/browse/EZP-27578)
- ~disable non translatable fields when in translation mode~ (see comments)
- ~write unit tests~ (will be taken care of later)
- ~write behat tests (see below)~ (will be taken care of later)

### DIFF
--- a/bundle/Resources/views/Content/form_fields.html.twig
+++ b/bundle/Resources/views/Content/form_fields.html.twig
@@ -1,0 +1,39 @@
+{% use 'form_div_layout.html.twig' %}
+
+{% block form_errors %}
+    {%- for error in errors if errors|length > 0 -%}
+        <em class="ez-field-edit-error">{{ error.message }}</em>
+    {%- endfor -%}
+{% endblock form_errors %}
+
+{% block form_label %}
+    {%- if form.value is defined -%}
+        {% set label_attr = label_attr|merge({'for': form.value.vars.id}) %}
+    {%- endif %}
+    {%- if compound is same as(false) -%}
+        {% set label_attr = label_attr|merge({'class': 'ez-sub-field-name'}) %}
+    {%- endif %}
+    {{- parent('form_label') -}}
+{% endblock form_label %}
+
+{%- block form_widget_compound -%}
+    <fieldset>
+        {% for child in form %}
+            <div class="ez-sub-field ez-sub-field-{{ child.vars.name }}">
+                <div class="ez-sub-field-text-zone">
+                    {{- form_label(child) -}}
+                    {{- form_errors(child) -}}
+                </div>
+                <div class="ez-sub-field-ui">
+                    {{- form_widget(child) -}}
+                </div>
+            </div>
+        {% endfor %}
+        {{- form_rest(form) -}}
+    </fieldset>
+{%- endblock form_widget_compound -%}
+
+{%- block number_widget -%}
+    {%- set type = type|default('number') -%}
+    {{ parent('number_widget') }}
+{%- endblock number_widget -%}
--- a/features/Context/FieldTypeFormContext.php
+++ b/features/Context/FieldTypeFormContext.php
@@ -69,13 +69,13 @@ final class FieldTypeFormContext extends RawMinkContext implements SnippetAccept
     }
 
     /**
-     * @Then /^the edit form should contain a fieldset named after the field definition$/
+     * @Then /^the edit form should contain an identifiable widget for that field definition$/
      */
     public function theEditFormShouldContainAFieldsetNamedAfterTheFieldDefinition()
     {
         $this->assertSession()->elementTextContains(
             'css',
-            sprintf('div.ezfield-identifier-%s fieldset legend', self::$fieldIdentifier),
+            'div.ez-field-edit div.ez-field-edit-text-zone label',
             'Field'
         );
     }
@@ -88,8 +88,7 @@ final class FieldTypeFormContext extends RawMinkContext implements SnippetAccept
         $this->assertSession()->elementExists(
             'css',
             sprintf(
-                'div.ezfield-identifier-%s fieldset input[type=%s]',
-                self::$fieldIdentifier,
+                'div.ez-field-edit div.ez-field-edit-ui input[type=%s]',
                 $inputType
             )
         );
@@ -104,7 +103,7 @@ final class FieldTypeFormContext extends RawMinkContext implements SnippetAccept
 
         $inputNodeElements = $this->getSession()->getPage()->findAll(
             'css',
-            sprintf('div.ezfield-identifier-%s fieldset input', self::$fieldIdentifier)
+            'div.ez-field-edit-ezuser div.ez-field-edit-ui fieldset div.ez-sub-field div.ez-sub-field-ui input'
         );
 
         /** @var NodeElement $nodeElement */
@@ -153,7 +152,7 @@ final class FieldTypeFormContext extends RawMinkContext implements SnippetAccept
     {
         $inputNodeElements = $this->getSession()->getPage()->findAll(
             'css',
-            sprintf('div.ezfield-identifier-%s fieldset input', self::$fieldIdentifier)
+            'div.ez-field-edit div.ez-field-edit-ui input'
         );
         Assertion::assertNotEmpty($inputNodeElements, 'The input field is not marked as required');
         foreach ($inputNodeElements as $inputNodeElement) {
--- a/lib/FieldType/Mapper/DateFormMapper.php
+++ b/lib/FieldType/Mapper/DateFormMapper.php
@@ -9,12 +9,20 @@
 namespace EzSystems\RepositoryForms\FieldType\Mapper;
 
 use eZ\Publish\Core\FieldType\Date\Type;
+use EzSystems\RepositoryForms\Data\Content\FieldData;
 use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\DataTransformer\DateValueTransformer;
 use EzSystems\RepositoryForms\FieldType\FieldDefinitionFormMapperInterface;
+use EzSystems\RepositoryForms\FieldType\FieldValueFormMapperInterface;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\DateType;
 use Symfony\Component\Form\FormInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class DateFormMapper implements FieldDefinitionFormMapperInterface
+/**
+ * FormMapper for ezdate FieldType.
+ */
+class DateFormMapper implements FieldDefinitionFormMapperInterface, FieldValueFormMapperInterface
 {
     public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
     {
@@ -35,5 +43,37 @@ class DateFormMapper implements FieldDefinitionFormMapperInterface
                     'translation_domain' => 'ezrepoforms_content_type',
                 ]
             );
+    }
+
+    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
+    {
+        $fieldDefinition = $data->fieldDefinition;
+        $formConfig = $fieldForm->getConfig();
+
+        $fieldForm
+            ->add(
+                $formConfig->getFormFactory()->createBuilder()
+                    ->create(
+                        'value',
+                        DateType::class,
+                        [
+                            'input' => 'datetime',
+                            'widget' => 'single_text',
+                            'required' => $fieldDefinition->isRequired,
+                            'label' => $fieldDefinition->getName($formConfig->getOption('languageCode')),
+                        ]
+                    )
+                    ->addModelTransformer(new DateValueTransformer())
+                    ->setAutoInitialize(false)
+                    ->getForm()
+            );
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setDefaults([
+                'translation_domain' => 'ezrepoforms_content_type',
+            ]);
     }
 }
--- a/lib/FieldType/Mapper/DateTimeFormMapper.php
+++ b/lib/FieldType/Mapper/DateTimeFormMapper.php
@@ -9,15 +9,22 @@
 namespace EzSystems\RepositoryForms\FieldType\Mapper;
 
 use eZ\Publish\Core\FieldType\DateAndTime\Type;
+use EzSystems\RepositoryForms\Data\Content\FieldData;
 use EzSystems\RepositoryForms\Data\FieldDefinitionData;
+use EzSystems\RepositoryForms\FieldType\DataTransformer\DateTimeValueTransformer;
 use EzSystems\RepositoryForms\FieldType\FieldDefinitionFormMapperInterface;
+use EzSystems\RepositoryForms\FieldType\FieldValueFormMapperInterface;
 use EzSystems\RepositoryForms\Form\Type\DateTimeIntervalType;
+use EzSystems\RepositoryForms\Form\Type\FieldValue\DateTimeLocalType;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class DateTimeFormMapper implements FieldDefinitionFormMapperInterface
+/**
+ * FormMapper for ezdatetime FieldType.
+ */
+class DateTimeFormMapper implements FieldDefinitionFormMapperInterface, FieldValueFormMapperInterface
 {
     public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
     {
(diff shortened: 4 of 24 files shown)


webhdx:
FYI: Fixed according to @ezrobot report and force pushed.

lib/FieldType/Mapper/DateTimeFormMapper.php
```diff
@@ -9,45 +9,109 @@
...
-                'required' => true,
-                'property_path' => 'fieldSettings[defaultType]',
-                'label' => 'field_definition.ezdatetime.default_type',
-            ])
-            ->add('dateInterval', DateTimeIntervalType::class, [
-                'required' => false,
-                'property_path' => 'fieldSettings[dateInterval]',
-                'label' => 'field_definition.ezdatetime.date_interval',
-            ]);
+            ->add(
```
dpobel: for next patches, it would be nice to not reindent code you don't really change, this makes big patches and then hard to review patches

lib/FieldType/Mapper/DateTimeFormMapper.php
```diff
@@ -9,45 +9,109 @@
...
+        $fieldForm
+            ->add(
+                $formConfig->getFormFactory()->createBuilder()
+                    ->create(
+                        'value',
+                        DateTimeType::class,
+                        [
+                            'input' => 'datetime',
+                            'date_widget' => 'single_text',
+                            'time_widget' => 'single_text',
```
dpobel: will that generate two inputs (`date` + `time`) or one `datetime` input ?
btw, in https://github.com/ezsystems/hybrid-platform-ui/pull/65/files#diff-46b5c70bde48cbbdc0324e8043b3be5cR378, it's specified this should be a `datetime-local` input (that does not seem to be possible out of the box according to https://symfony.com/doc/current/reference/forms/types/datetime.html)

lib/FieldType/Mapper/FloatFormMapper.php
```diff
@@ -61,7 +83,42 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
+        $formConfig = $fieldForm->getConfig();
+        $names = $fieldDefinition->getNames();
+        $label = $fieldDefinition->getName($formConfig->getOption('languageCode')) ?: reset($names);
+
+        $fieldForm
+            ->add(
+                $formConfig->getFormFactory()->createBuilder()
+                    ->create(
+                        'value',
+                        NumberType::class,
```
dpobel: according to http://symfony.com/doc/current/reference/forms/types/number.html this renders a `text` input (?!?) Also, `min`, `max` and `step` are missing (which indeed is *normal* if that's really rendered as a `text` input)

lib/FieldType/Mapper/ISBNFormMapper.php
```diff
@@ -55,7 +75,42 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
+
+        $fieldForm
+            ->add(
+                $formConfig->getFormFactory()->createBuilder()
+                    ->create(
+                        'value',
+                        TextType::class,
+                        [
+                            'required' => $fieldDefinition->isRequired,
+                            'label' => $label,
```
dpobel: no `pattern` (https://github.com/ezsystems/hybrid-platform-ui/pull/65/files#diff-46b5c70bde48cbbdc0324e8043b3be5cR726) ? no way to set it with SF TextType ?

lib/FieldType/Mapper/IntegerFormMapper.php
```diff
@@ -61,7 +82,53 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
+        $fieldForm
+            ->add(
+                $formConfig->getFormFactory()->createBuilder()
+                    ->create(
+                        'value',
+                        IntegerType::class,
+                        [
+                            'required' => $fieldDefinition->isRequired,
+                            'label' => $label,
+                            'attr' => $attributes,
```
dpobel: does `IntegerType` force the `step` attribute to `1` ?

lib/FieldType/Mapper/TextLineFormMapper.php
```diff
@@ -72,9 +100,16 @@ public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
                     ->create(
                         'value',
                         TextType::class,
-                        ['required' => $fieldDefinition->isRequired, 'label' => $label]
+                        [
+                            'required' => $fieldDefinition->isRequired,
+                            'label' => $label,
```
dpobel: missing `pattern` as well to take `minlength` and `maxlength` into account

lib/Form/Type/FieldValue/UrlType.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add(
+                'link',
+                BaseUrlType::class,
+                [
+                    'required' => $options['required'],
+                    'label' => 'URL',
```
dpobel: those labels should be translatable

lib/FieldType/Mapper/DateFormMapper.php
```diff
@@ -36,4 +47,40 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
+    /**
+     * @param FormInterface $fieldForm
+     * @param FieldData $data
+     */
+    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
+    {
+        $fieldDefinition = $data->fieldDefinition;
+        $formConfig = $fieldForm->getConfig();
+        $names = $fieldDefinition->getNames();
+        $label = $fieldDefinition->getName($formConfig->getOption('languageCode')) ?: reset($names);
```
dpobel: it's the same for all mappers, which makes me think this should be extracted and reused instead of duplicated in every mapper

lib/FieldType/Mapper/FloatFormMapper.php
webhdx: Take a look at that: https://github.com/symfony/symfony/blob/2.1/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L160

What do you think?

lib/FieldType/Mapper/ISBNFormMapper.php
webhdx: I believe it can be added under `attr` option. I will look into that.

lib/FieldType/Mapper/IntegerFormMapper.php
webhdx: @dpobel Isn't it the default value when `step` attribute is not provided? I can enforce it though, no problem.

lib/FieldType/Mapper/FloatFormMapper.php
dpobel: > What do you think?

I think that's a good example of a very bad comment :)
Joke aside, maybe at some point there were some issues but according to http://caniuse.com/#feat=input-number and my personal test page http://vrac.pwet.fr/input.html type=number input seems to work well for both Integer and Float

See also https://github.com/symfony/symfony/issues/8106

lib/FieldType/Mapper/IntegerFormMapper.php
dpobel: you are right, as long as IntegerType renders a type="number" input that's fine

lib/FieldType/Mapper/FloatFormMapper.php
webhdx: @dpobel Keep in mind I'm not a frontend developer so I'm usually behind current state of HTML browser support. Fortunately it's can be easily fixed on our side.

webhdx:
- Fixed all issues according to @dpobel's review. 
- Rebased to master as #133 has been already merged.

I talked with @bdunogier and we decided to keep unit tests out of this PR as it's already large and there are many more fields to come. After finishing all fieldtypes integration I will move to writing unit tests.

BTW. @bdunogier you can always review commit-by-commit if it makes it easier for you. I reverted code reformatting as you requested.

bdunogier:
Could you refine the messages so that they refer to the actual change ? I haven't looked at all of them yet, but 612436f is very generic, while it seems to really do something on its own :)

lib/FieldType/Mapper/TextLineFormMapper.php
```diff
@@ -14,23 +14,31 @@
...
+    /**
+     * @param FieldTypeService $fieldTypeService
+     */
     public function __construct(FieldTypeService $fieldTypeService)
     {
         $this->fieldTypeService = $fieldTypeService;
     }
 
+    /**
+     * @param FormInterface $fieldDefinitionForm
```
bdunogier: To be honest, I wouldn't bother with those. They're not adding anything to what the method already says (parameters type and name), and they only repeat the interface.

Some of us choose to use @inheritdoc for that, to show that the phpdoc wasn't forgotten. I prefer not to. Either way, no paraphrasing the prototypes. The best code is the one we don't have to write ;)

lib/FieldType/Mapper/TextLineFormMapper.php
```diff
@@ -59,20 +67,33 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
                 $formConfig->getFormFactory()->createBuilder()
                     ->create(
                         'value',
                         TextType::class,
-                        ['required' => $fieldDefinition->isRequired, 'label' => $label]
+                        [
+                            'required' => $fieldDefinition->isRequired,
+                            'label' => $fieldDefinition->getName($formConfig->getOption('languageCode')),
+                            'attr' => [
+                                'pattern' => sprintf(
```
bdunogier: Maybe this could go to a private method, just for readability ?

lib/FieldType/Mapper/TextLineFormMapper.php
```diff
@@ -59,20 +67,33 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
     }
 
+    /**
+     * @param FormInterface $fieldForm
+     * @param FieldData $data
+     */
     public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
     {
         $fieldDefinition = $data->fieldDefinition;
+        $validatorConfiguration = $fieldDefinition->getValidatorConfiguration();
```
bdunogier: If you create a private method below, I think you can call `getValidatorConfiguration()` directly from the method call and drop this line.

lib/FieldType/Mapper/DateFormMapper.php
```diff
@@ -9,13 +9,25 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class DateFormMapper implements FieldDefinitionFormMapperInterface
+/**
+ * FormMapper for ezdate FieldType.
+ */
+class DateFormMapper implements FieldDefinitionFormMapperInterface, FieldValueFormMapperInterface
 {
+    /**
+     * @param FormInterface $fieldDefinitionForm
```
bdunogier: Same as for Text. No need to repeat the phpdoc of interfaces.

lib/FieldType/Mapper/FloatFormMapper.php
```diff
@@ -61,7 +71,54 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
+    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
+    {
+        $fieldDefinition = $data->fieldDefinition;
+        $validatorConfiguration = $fieldDefinition->getValidatorConfiguration();
+        $formConfig = $fieldForm->getConfig();
+        $attributes = [
+            'step' => 1,
+        ];
+
+        if (null !== $validatorConfiguration['FloatValueValidator']['minFloatValue']) {
```
bdunogier: I'd move that to a private method, like `mapValidatorConfiguration($validatorConfiguration)`.

lib/FieldType/Mapper/FloatFormMapper.php
```diff
@@ -61,7 +71,54 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
     /**
+     * @param FormInterface $fieldForm
+     * @param FieldData $data
+     */
+    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
+    {
+        $fieldDefinition = $data->fieldDefinition;
+        $validatorConfiguration = $fieldDefinition->getValidatorConfiguration();
+        $formConfig = $fieldForm->getConfig();
+        $attributes = [
```
bdunogier: This would fit on one line :)

lib/FieldType/Mapper/IntegerFormMapper.php
```diff
@@ -61,7 +71,54 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
...
+    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
+    {
+        $fieldDefinition = $data->fieldDefinition;
+        $validatorConfiguration = $fieldDefinition->getValidatorConfiguration();
+        $formConfig = $fieldForm->getConfig();
+        $attributes = [
+            'step' => 1,
+        ];
+
+        if (null !== $validatorConfiguration['IntegerValueValidator']['minIntegerValue']) {
```
bdunogier: Same remark about the private method.

lib/FieldType/Mapper/TextLineFormMapper.php
dpobel: also, if the field definition has no length constraint, this will generate `pattern=".{,}"` which is bit weird (and I'm unsure of the consequences in browsers)

bundle/Resources/views/Content/form_fields.html.twig
```diff
@@ -0,0 +1,42 @@
...
+    {%- if compound is same as(false) -%}
+        {% set label_attr = label_attr|merge({'class': 'ez-sub-field-name'}) %}
+    {%- endif %}
+    {{- parent('form_label') -}}
+{% endblock form_label %}
+
+{%- block form_widget_compound -%}
+    <fieldset>
+        {%- if form.parent is empty -%}
+            {{ form_errors(form) }}
```
dpobel: validation errors are not supposed to appear here but in `div.ez-field-edit-text-zone` (as written in https://github.com/ezsystems/repository-forms/pull/134/files#diff-6972ff935106f8fc8461d54811ad45caR21). That said, if we have errors dedicated to *sub fields*, it should appear in the corresponding `div.ez-sub-field-text-zone`

webhdx:
@bdunogier @dpobel Thanks for the feedback. Fixed according to your review.

bdunogier:
Actually, the Travis failure is a real one, I think. It is about the content creation form, and it fails because the expected field markup isn't found. Since this pull-request changes it, you're gonna have to update the behat sentence's method.

Once it's green we'll merge it.

webhdx:
@bdunogier Tests are fixed but they require to be revisited after we are closer to getting whole content editing done. Not only they miss a lot of stuff but there are some parts that are now poorly checked i.e.: we can't use CSS selectors to access `.ez-field-edit` container element of a particular field using its identifier. This is due to insufficient markup but might be possible to workaround by using XPath. I'm not Behat expert tho. Anyway, tests are still logically valid and passing.

features/Context/FieldTypeFormContext.php
```diff
@@ -69,13 +69,13 @@ public function iEditOrCreateContentOfThisType()
     }
 
     /**
-     * @Then /^the edit form should contain a fieldset named after the field definition$/
+     * @Then /^the edit form should contain a container div labelled after the field definition$/
```
bdunogier: If we wanted to be a bit more future proof, we could find a phrasing that doesn't refer to the technical elements. 
What about `Then the edit form should contain an identifiable widget for that field definition`.

The idea is to say that it should show up. How we check it is up to us. The "identifiable" part is to make sure that we can identify it from a frontend api perspective.

What do you think ?

webhdx: @bdunogier Seems reasonable. Fixed and pushed.

PR #136, opened 2017-07-21
EZP-27659 Subtree and Node policy limitations can't be viewed

glye:
> Fix https://jira.ez.no/browse/EZP-27659
> QA

Existing node/subtree limitations are not displayed when editing a policy. This ensures they are visible.

Fixes:
- [x] Subtreelimitation saves paths (legacy), UDW wants location IDs. Add conversion in the transformer.
- [x] The form (hidden input text line) needs the IDs comma separated, so that's what the value is, but we have to split them in the template to render their names
- [x] The IDs are locations not content, so render them correctly as locations
- [x] ~Preselect node/subtree limitations in UDW~ Skipped, as that is a separate issue

### DIFF
--- a/bundle/Resources/views/Limitation/udw_limitation_value.html.twig
+++ b/bundle/Resources/views/Limitation/udw_limitation_value.html.twig
@@ -9,9 +9,9 @@
 >{{ "role.policy.limitation.location.udw_button"|trans({}, "ezrepoforms_role") }}</button>
 <div>
     <ul id="{{form.limitationValues.vars.id}}-selected-location">
-        {% for limitationValue in form.limitationValues.vars.value %}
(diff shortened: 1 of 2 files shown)


glye:
> Preselect node/subtree limitations in UDW

@dpobel Does V1 UDW support this at all? I mean not a browsing start location, but that certain items are already in the "Confirmed items" list when UDW opens.

If it can't do this, we'd either need to add delete buttons for items in the selected item list in policy edit, or skip it alltogether and just accept that whatever you select in UDW overwrites what you had before. (Perhaps suboptimal, but will be replaced in V2 anyway)

glye:
Thanks, sent to QA.

PR #137, opened 2017-07-21
Fix EZP-27513: Disable checking uniqueness of role identifier for null

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27512

# Description

`UniqueRoleIdentifier` validator should not be executed when `identifier` is `null`, instead error should be raised by `NotBlank` validator (already configured https://github.com/ezsystems/repository-forms/blob/master/bundle/Resources/config/validation.yml#L68). 

# Steps to reproduce
1. Create a new role
2. Set the name as a space
3. Save the role
4. Error `Failed to load 'http://.../pjax/role/update/8` instead of `Invalid form data` is displayed to user

### DIFF
--- a/lib/Validator/Constraints/UniqueRoleIdentifierValidator.php
+++ b/lib/Validator/Constraints/UniqueRoleIdentifierValidator.php
@@ -36,7 +36,7 @@ class UniqueRoleIdentifierValidator extends ConstraintValidator
      */
     public function validate($value, Constraint $constraint)
     {
-        if (!$value instanceof RoleData) {
+        if (!$value instanceof RoleData || $value->identifier === null) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


adamwojs:
> Will the space be converted to null along the way, so it ends up being the same thing?

@glye Yes, it is. A more detailed answer to the question "Why validator is executed against `null` value instead of `' '` ?" is: 

1. The `Identifier` field has enabled by default a `trim` option https://symfony.com/doc/2.8/reference/forms/types/text.html#trim, which is handled by https://github.com/symfony/form/blob/2.8/Extension/Core/EventListener/TrimListener.php#L34  In our case (string with single space), we will get an empty string as result, when `FormEvents::PRE_SUBMIT` event is dispatched (https://github.com/symfony/form/blob/2.8/Form.php#L558).

2. The empty string is "converted" to `null`, when submitted data are normalized (transformed from view data into norm data). This happens in this line of code: https://github.com/symfony/form/blob/2.8/Form.php#L1185, when there is not any other view transformer attached to form field.

adamwojs:
_Note: I've changed the commit message / PR title according to @bdunogier comment in https://github.com/ezsystems/repository-forms/pull/138_

bdunogier:
And merged. Thank you @adamwojs.

PR #138, opened 2017-07-21
Fix EZP-27690: Disable checking uniqueness of content type identifier for null

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27690

# Description
As same as https://github.com/ezsystems/repository-forms/pull/137: `UniqueContentTypeIdentifier` validator should not be executed when identifier is `null`, instead, an error should be raised by `NotBlank` validator (already configured https://github.com/ezsystems/repository-forms/blob/master/bundle/Resources/config/validation.yml#L6).

# Steps to reproduce
1.  Create a new content type
2. Set the identifier as a space
3. Save the content type
4.  Notification contains text
`Failed to load 'http://192.168.56.102/pjax/contenttype/update/2/eng-GB'`
instead of a regular validation error message. Error labels aren't displayed under form fields.

### DIFF
--- a/lib/Validator/Constraints/UniqueContentTypeIdentifierValidator.php
+++ b/lib/Validator/Constraints/UniqueContentTypeIdentifierValidator.php
@@ -41,7 +41,7 @@ class UniqueContentTypeIdentifierValidator extends ConstraintValidator
      */
     public function validate($value, Constraint $constraint)
     {
-        if (!$value instanceof ContentTypeData) {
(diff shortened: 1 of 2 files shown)


adamwojs:
@bdunogier I'm glad that you paying attention 😉  The length of a message is now 78 characters.

bdunogier:
I have rephrased it a tiny little bit during merge. Thank you @adamwojs.

adamwojs:
Yes, I've moved this issue to QA.

PR #139, opened 2017-07-27
[WIP] Custom ez-field-edit element (EZP-27576)

bdunogier:
> [EZP-27576](http://jira.ez.no/browse/EZP-27576)

Wraps field edit forms into a customizable (not yet) `<ez-field-edit />` element:

```html
<ez-field-edit fieldtype="ezstring" field-settings="{}" validator-configuration='{"StringLengthValidator":{"maxStringLength":20,"minStringLength":10}}'>
  <!-- markup for TextLine -->
</ez-field-edit>
```

This is implemented in `@EzSystemsRepositoryFormsBundle/Content/content_form.html.twig`, by serializing the `FieldDefinition` into attributes.

### TODO
- [ ] Make the custom element customizable
- [ ] Figure out if escaping should be modified, or if it's okay

### DIFF
--- a/bundle/Resources/views/Content/content_form.html.twig
+++ b/bundle/Resources/views/Content/content_form.html.twig
@@ -1,37 +1,46 @@
 {% macro display_form(form) %}
+    {# Should be a property we get from the FieldMapper #}
+    {% set fieldEditTag = 'ez-field-edit' %}
     {% form_theme form with ['EzSystemsRepositoryFormsBundle:Content:form_fields.html.twig', _self] %}
 
     {{ form_start(form) }}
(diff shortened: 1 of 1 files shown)


bdunogier:
Rebased on master now that fieldtypes are merged.

webhdx:
@dpobel How about putting field identifier/name in there too? It would be very helpful for behat tests and probably for writting custom JS as well. Currently field identifier is only applied as `id` attribute on `<input>` element. This is okay if you only need to access the input but doesn't allow to easily access `<label>` or wrapping element using CSS selectors.

PR #143, opened 2017-08-11
EZP-27762 content/edit routing conflict with legacy

glye:
> Fix https://jira.ez.no/browse/EZP-27762
> Ready for review

Ensure our route doesn't conflict with legacy content edit, by changing `/content/edit/...` to `/content/edit/draft/...`.

Debatable: Renaming the route from `ez_content_edit` to `ez_content_draft_edit`, as I've done. The fix is smaller if we don't.

### DIFF
--- a/bundle/Resources/config/routing.yml
+++ b/bundle/Resources/config/routing.yml
@@ -3,8 +3,8 @@ ez_content_create_no_draft:
     defaults:
         _controller: ez_content_edit:createWithoutDraftAction
 
-ez_content_edit:
-    path: /content/edit/{contentId}/{versionNo}/{language}
+ez_content_draft_edit:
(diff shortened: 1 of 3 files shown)


andrerom:
NOTE: One alternative here is to pick a different verb then `edit`, could be `update` to closer adhere to CRUD terminology for instance.

bdunogier:
> could be `update` to closer adhere to CRUD terminology

Not bad. But the downside is that this resource doesn't only update, it also publishes (if you ask it to). `content/edit/draft` sounds alright to me.

bdunogier:
Also, do we care about the BC break (route change) ?

andrerom:
> Also, do we care about the BC break (route change) ?

Which bc break? This was reverted from stable branch before 1.11 release, so it's only in master.

PR #144, opened 2017-08-24
EZP-27793: fix updating ContentType with required value

wizhippo:
This allows ContentTypes to have an empty value even when marked as required.

My interpretation of 'required' applies to the the ContentField not the Type.

https://jira.ez.no/browse/EZP-27793

### DIFF
--- a/lib/Validator/Constraints/FieldValueValidator.php
+++ b/lib/Validator/Constraints/FieldValueValidator.php
@@ -13,6 +13,7 @@ namespace EzSystems\RepositoryForms\Validator\Constraints;
 use eZ\Publish\API\Repository\Values\ValueObject;
 use eZ\Publish\Core\FieldType\ValidationError;
 use EzSystems\RepositoryForms\Data\Content\FieldData;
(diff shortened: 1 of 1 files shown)


glye:
"Required" should refer to the field value, not the field type, yes. I confirm the behaviour described in the issue. I'm surprised that the `FieldValueValidator` runs for field type definitions, as the fix indicates.

andrerom:
@glye WDYT?

andrerom:
> though as André says it should not go into master.

I assumed this might affect 1.11 _(1.9 here)_, but I see @wizhippo reports it as issue in dev-master only. Anyone know when this issue was introduced?

glye:
> Anyone know when this issue was introduced?

It seems to be from https://github.com/ezsystems/repository-forms/commit/0deb748d251c093719e8dba88311e1b39d88d2e3 at July 12th, in 1.9.0 (based on an older commit).

andrerom:
so 1.9 it is then, @wizhippo can you swap branch?

wizhippo:
@andrerom That should do it

andrerom:
Thanks, merged :)

PR #145, opened 2017-08-25
EZP-27646: Implement editing support for User FieldType

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-27646
> JIRA (epic): https://jira.ez.no/browse/EZP-26700

# Description
* `ezuser` [EZP-27646: Implement editing support for User FieldType](https://jira.ez.no/browse/EZP-27646) 

# Remarks
I had to implement completely different infrastructure for editing Users due to how kernel manages users. My implementation introduces separate action dispatcher with new event names, separated form processors, types, data classes etc.

# TODO
- [ ] document BC breaks
- [ ] check Behat tests
- [ ] check unit tests

### DIFF
--- a/bundle/Controller/UserController.php
+++ b/bundle/Controller/UserController.php
@@ -1,0 +1,181 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace EzSystems\RepositoryFormsBundle\Controller;
+
+use eZ\Bundle\EzPublishCoreBundle\Controller;
+use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
+use eZ\Publish\API\Repository\LanguageService;
+use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\UserService;
+use eZ\Publish\Core\Base\Exceptions\BadStateException;
+use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
+use eZ\Publish\Core\Base\Exceptions\InvalidArgumentType;
+use EzSystems\RepositoryForms\Data\Mapper\UserCreateMapper;
+use EzSystems\RepositoryForms\Data\Mapper\UserUpdateMapper;
+use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
+use EzSystems\RepositoryForms\Form\Type\User\UserCreateType;
+use EzSystems\RepositoryForms\Form\Type\User\UserUpdateType;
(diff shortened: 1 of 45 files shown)


bundle/Controller/UserController.php
```diff
@@ -0,0 +1,168 @@
...
+     * Displays and processes a user creation form.
+     *
+     * @param string $contentTypeIdentifier ContentType id to create
+     * @param string $language Language code to create the content in (eng-GB, ger-DE, ...))
+     * @param int $parentLocationId Location the content should be a child of
+     * @param Request $request
+     *
+     * @return UserCreateView|Response
+     *
+     * @throws InvalidArgumentType
```
bdunogier: Ain't that a lot of exceptions for such a little method ? :)
I'm wondering if we shouldn't catch and group most of these into exceptions that make sense from the standpoint of a controller.

bundle/Resources/config/services.yml
```diff
@@ -31,21 +29,26 @@ parameters:
...
     ezrepoforms.action_dispatcher.content.class: EzSystems\RepositoryForms\Form\ActionDispatcher\ContentDispatcher
+    ezrepoforms.action_dispatcher.user.class: EzSystems\RepositoryForms\Form\ActionDispatcher\UserDispatcher
     ezrepoforms.form_processor.content_type.class: EzSystems\RepositoryForms\Form\Processor\ContentTypeFormProcessor
     ezrepoforms.form_processor.content_type.options.redirect_route_after_publish: ~
     ezrepoforms.form_processor.content_type.options:
         redirectRouteAfterPublish: "%ezrepoforms.form_processor.content_type.options.redirect_route_after_publish%"
     ezrepoforms.form_processor.content_type_group.class: EzSystems\RepositoryForms\Form\Processor\ContentTypeGroupFormProcessor
     ezrepoforms.form_processor.content.class: EzSystems\RepositoryForms\Form\Processor\ContentFormProcessor
-    ezrepoforms.form_processor.user_register.class: EzSystems\RepositoryForms\Form\Processor\UserRegisterFormProcessor
+    ezrepoforms.form_processor.user_create.class: EzSystems\RepositoryForms\Form\Processor\User\UserCreateFormProcessor
```
bdunogier: One remark about the services definitions: what about adding a new file for the user stuff, that uses the newer conventions (no `%%` placeholders for classes names, for instance) ? At least we wouldn't have to change those later on.

It would also help keep the files reasonable in size, and that can't hurt :)

webhdx:
Hey @bdunogier. I really like "the master plan" and will happily work on this further more. With current setup there is still a lot of flexibility as developers are able to inject their templates, extend editing types, handle actions differently using custom processors etc. We use that a lot with AdminUI. I like the idea but for the time being I think it's just best to extend what we have. This was my goal with this PR.
Maybe in the beginning of the next year there will be more time to do improvements or I will handle it as a Knowledge Friday task.

bdunogier:
> !! Do not merge until #140 is merged. It has to be rebased onto master first !!

You can remove this, since #140 has been merged.

bdunogier:
No problem @webhdx. I don't imply that this should be applied to this PR anyway, the change is large enough as it is. If it helps you better understand what the intention was, it was good enough.

On the long term, it's more about building an architecture we can scale from. The rest is a set of details :)

PR #149, opened 2017-08-29
EZP-27633: Implement editing support for Country FieldType

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-27633
> **!! Please do not merge until #140 is merged. I have to rebase on master and skip duplicated commit first !!**

# Description
* `ezcountry` [EZP-27633: Implement editing support for Country FieldType](https://jira.ez.no/browse/EZP-27633) 

Please note it isn't prepared for Content translation yet.

### DIFF
--- a/doc/bc/changes-1.10.md
+++ b/doc/bc/changes-1.10.md
@@ -16,6 +16,12 @@ Changes affecting version compatibility with former or future versions.
   How it might affect your code: Don't rely on `TranslationHelper` as the same functionality is now achieved by calling `getNames()` directly on the object.
 
 ## Deprecations
+- EZP-27633: Implement editing support for Country FieldType
+
+  What changed: `\EzSystems\RepositoryForms\FieldType\DataTransformer\CountryValueTransformer` is deprecated. It will be removed in 2.0.
+
+  How it might affect your code: BC is preserved for the time being but it's advised to use `\EzSystems\RepositoryForms\FieldType\DataTransformer\MultipleCountryValueTransformer` which can be done without any effort besides class name change.
+
  - The "ezrepoforms.user_register.view_templates_listener" service is deprecated since 1.10 and will be removed in 2.0. Use "ezrepoforms.view_templates_listener" instead
    
 ## Removed features
--- a/lib/FieldType/DataTransformer/MultipleCountryValueTransformer.php
+++ b/lib/FieldType/DataTransformer/MultipleCountryValueTransformer.php
@@ -1,0 +1,55 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\FieldType\DataTransformer;
+
+use eZ\Publish\Core\FieldType\Country\Value;
+use Symfony\Component\Form\DataTransformerInterface;
+
+/**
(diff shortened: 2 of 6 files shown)


doc/bc/changes-1.10.md
```diff
@@ -0,0 +1,14 @@
+# Backwards compatibility changes
+
+Changes affecting version compatibility with former or future versions.
+
+## Changes
+- EZP-27633: Implement editing support for Country FieldType
+
+  What changed: `EzSystems\RepositoryForms\FieldType\DataTransformer\CountryValueTransformer` was renamed to `MultipleCountriesValueTransformer`.
```
dpobel: that's a real BC break and this one can be easily avoided. For instance, instead of a creating a new transformer, you could add a `multiple` parameter to the constructor `CountryValueTransformer` (which a default value for BC reason) and then change `transform` and `reverseTransform` to take it into account.

Another option, if you really want to have 2 transformers is to keep a deprecated `CountryValueTransformer` it can either remain as it is now or it can just extend `MultipleCountriesValueTransformer`.

webhdx:
@dpobel Fixed. I opted for the first approach i.e. additional constructor argument to avoid too many classes. Someone can remove @ezrobot review as I don't have permissions. I've fixed it already.

lib/FieldType/Mapper/CountryFormMapper.php
```diff
@@ -8,24 +8,22 @@
...
 {
-    /**
-     * @var array Array of countries from ezpublish.fieldType.ezcountry.data
-     */
+    /** @var array Array of countries from ezpublish.fieldType.ezcountry.data */
     protected $countriesInfo;
 
-    /**
-     * @param array $countriesInfo
-     */
```
glye: Why was this doc removed?

andrerom: Might be because the type hint and param name is already in the php code, so as long as phpdoc can't add more info than wheat the code does some people tend to prefer removing it.

glye: Ack, would prefer adding a description then instead, unless that's not really needed.

andrerom: true, but it's a ctor, it's usually self describing

webhdx: Actually I haven't intended to remove that phpdoc, removed it by mistake. Fixed it :)

webhdx:
@lserwatka  This one can be merged.

lib/FieldType/Mapper/CountryFormMapper.php
bdunogier: Actually I agreed with @andrerom. Without added information, it's mostly an extra risk that we end-up with mismatching info whenever something changes. But let's not touch that now.

lib/FieldType/DataTransformer/MultipleCountryValueTransformer.php
```diff
@@ -0,0 +1,55 @@
...
+use Symfony\Component\Form\DataTransformerInterface;
+
+/**
+ * DataTransformer for Country\Value to be used with form type handling multiple selections.
+ * Needed to display the form field correctly and transform it back to an appropriate value object.
+ */
+class MultipleCountryValueTransformer implements DataTransformerInterface
+{
+    /**
+     * @var array Array of countries from ezpublish.fieldType.ezcountry.data
```
bdunogier: I wouldn't specify where the value is taken from. The service container already has that info, and it kind of adds a two-way relationship between the class and the service definition.

PR #151, opened 2017-08-29
EZP-27628: Implement editing support for RichText FieldType

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-27628
> **!! Please do not merge until #140 is merged. I have to rebase on master and skip duplicated commit first !!**

# Description
* `ezrichtext` [EZP-27628: Implement editing support for RichText FieldType](https://jira.ez.no/browse/EZP-27628) 

It exposes `<textarea></textarea>` element accepting RichText's XML markup.

Please note it isn't prepared for Content translation yet. It also requires JS enrichment done in Hybrid UI.

### DIFF
--- a/doc/bc/changes-1.10.md
+++ b/doc/bc/changes-1.10.md
@@ -3,6 +3,12 @@
 Changes affecting version compatibility with former or future versions.
 
 ## Changes
+- EZP-27628: Implement editing support for RichText FieldType
+
+  What changed: `ezrepoforms.field_type.form_mapper.ezrichtext` requires `ezpublish.api.service.field_type` and `ezpublish.fieldType.ezrichtext.converter.edit.xhtml5` dependencies.
+
+  How it might affect your code: If you are implementing a subclass, make sure you are injecting required services.
+  
 - EZP-27641: Implement editing support for Relation FieldType
 
   What changed: Removed `ezpublish.translation_helper` dependency from `ezrepoforms.field_type.form_mapper.abstractrelation` service.
--- a/lib/FieldType/DataTransformer/RichTextValueTransformer.php
+++ b/lib/FieldType/DataTransformer/RichTextValueTransformer.php
@@ -1,0 +1,63 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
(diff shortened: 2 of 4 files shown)


doc/bc/changes-1.10.md
```diff
@@ -0,0 +1,14 @@
+# Backwards compatibility changes
+
+Changes affecting version compatibility with former or future versions.
+
+## Changes
+- EZP-27628: Implement editing support for RichText FieldType
+
+  What changed: `ezrepoforms.field_type.form_mapper.ezrichtext` requries `ezpublish.api.service.field_type` dependency.
```
dpobel: `requires`

lib/FieldType/DataTransformer/RichTextValueTransformer.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @return string
+     */
+    public function transform($value)
+    {
+        if (!$value instanceof Value) {
+            return '';
+        }
+
+        return (string) $value;
```
dpobel: If I get that correctly, that means the textarea will be filled with the XML internal format (docbook) which won't be super usable to have a Rich Text editor :)
Instead, the textarea should be filled with `xhtml5edit` representation of the field value

lib/FieldType/DataTransformer/RichTextValueTransformer.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @return Value|null
+     */
+    public function reverseTransform($value)
+    {
+        if ($value === null || empty($value)) {
+            return $this->fieldType->getEmptyValue();
+        }
+
+        return $this->fieldType->fromHash(['xml' => $value]);
```
dpobel: and maybe same here, is it enough to accept an `xhtml5edit` document ?

webhdx: Funny fact - This is how it works in PlatformUI. UI is sending JSON object with `xml` and `xhtml5edit` properties but both contain the same markup (xhtml5edit). Kinda hackish but it is later picked up by a processor which is identifying the schema and converting to docbook. It works the way its implemented above as `fromHash()` is taking care of it.

lib/FieldType/DataTransformer/RichTextValueTransformer.php
webhdx: Oh, you are right. I wasn't aware we are using two different schemas. The fact we are using some processors doing work "behind the scenes" didn't helped either. Anyway, fixed it.

PR #152, opened 2017-08-29
EZP-27819: Template exception when editing policies with no node and subtree limitation

glye:
> Fixes https://jira.ez.no/browse/EZP-27819
> Pending QA

When editing limitations for a policy that can have Node and Subtree limitations, but doesn't, Twig throws an exception because it is trying to render a Location with an empty string as ID.

Fixing this by verifying we have an ID to render. Thanks to @Plopix!


glye:
@andrerom Yep, I've verified 1.7 is not affected, FT only.

glye:
Sent to QA.

PR #153, opened 2017-09-07
[WIP] EZP27628: Check tests on Travis

mnocon:
This PR is related to test failures that occurred in https://github.com/ezsystems/repository-forms/pull/149 - it checks whether the tests are stable on Travis in master branch.

It should not be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: php
 sudo: required
 
+env:
+  global:
+    # For behat tests
+    - COMPOSE_FILE="doc/docker-compose/base-prod.yml:doc/docker-compose/selenium.yml"
+
 matrix:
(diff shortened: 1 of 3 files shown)


alongosz:
It's related to [Travis upgrade of the default dist to Trusty](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status).
To solve that, I tried to use docker for behat as we do on other repos, but still something seems to be missing.
@mnocon for now your fix to use `dist: precise` on Travis seems to be the best, until we figure this out.

alongosz:
Closing in favor of #154

andrerom:
I guess prepare_behat.sh will need to be adopted to align with the one in kernel also, + maybe more env vars.

PR #154, opened 2017-09-07
[WIP] EZP-27860: Set Travis dist to "precise"

mnocon:
> JIRA: [EZP-27860](https://jira.ez.no/browse/EZP-27860)

# Description

In this repository we've been using the default dist on Travis , which used to be "precise". It has been changed recently [source](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status) (so that it default to "trusty" and it looks our behat tests are not working correctly there.

This PR changes the distribution back to the one we've been using to make sure the tests are passing. It should unblock some tickets, such as:
https://github.com/ezsystems/repository-forms/pull/140
https://github.com/ezsystems/repository-forms/pull/149

I'd treat this change as a quick fix so that the development is unblocked, I guess a deeper investigation is needed to find out why aren't the tests working on Trusty.

I'm marking it as WIP until the Travis build passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: php
 sudo: required
+dist: precise
 
 matrix:
   # mark as finished before allow_failures are run


alongosz:
Hmm, I wonder why we don't run Travis on `1.5` given [LTS supports `~1.5.0`](https://github.com/ezsystems/ezplatform/blob/v1.7.5/composer.json#L31).
I think [here](https://github.com/ezsystems/ezplatform/blob/v1.7.5/composer.json#L31) we should drop `1.1` - `1.4` and add both `1.5` and latest `1.9`. Or just replace it with regexp matching any stable branch `/^\d.\d+$/`.
WDYT @yannickroger?

andrerom:
@alongosz your comment was fixed after merge: https://github.com/ezsystems/repository-forms/commit/4a7d3d0e2d68844cb0be91d7d318a8e29ffe5cfa

PR #156, opened 2017-09-13
Removed getMock method and added class constants

MarioBlazek:
This PR resolves:
* removed deprecated `getMock` method
* tests updated with class constants
* fixed minor typos (buildVioloation -> buildViolation)


MarioBlazek:
@adamwojs is it actually necessary ? The whole PR will be squashed on merge, right ?

adamwojs:
@MarioBlazek No, it isn't necessary. This is only note for the future 😉

andrerom:
I missed the `UniqueLanguageCodeValidatorTest` addition on first look, what is it for? And seems like a separate concern so should ideally be separate commit, or if a good enough reason for it a separate commit and we'll create merge commit instead of squash*.

\* _This is why @adamwojs suggested separate commit, we don't always squash, if there are several authors or several topics handled in a PR we create merge commit_

MarioBlazek:
@adamwojs @andrerom should I add `UniqueLanguageCodeValidatorTest` to separate commit ?

andrerom:
> should I add UniqueLanguageCodeValidatorTest to separate commit ?

either own PR, or rebase this to have two commits (one for getMock and the rest, and one for adding new tests for UniqueLanguageCodeValidatorTest)

MarioBlazek:
@andrerom created #157 and removed `UniqueLanguageCodeValidatorTest` from this PR

andrerom:
Thanks :)

PR #158, opened 2017-09-19
EZP-27884: Create extensibility point for views overriding in content edit and draft create

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-27884

# Description
Changed return type of the controller's actions. It now returns `View` objects which are rendered by `ViewManager`. Templates are injected from parameters which can be easily configured. It uses the same principle as UserRegister.

### DIFF
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -14,6 +14,8 @@ use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo;
 use eZ\Publish\Core\Base\Exceptions\BadStateException;
+use EzSystems\RepositoryForms\Content\View\ContentCreateDraftView;
+use EzSystems\RepositoryForms\Content\View\ContentEditView;
 use EzSystems\RepositoryForms\Data\Content\CreateContentDraftData;
 use EzSystems\RepositoryForms\Data\Mapper\ContentCreateMapper;
 use EzSystems\RepositoryForms\Data\Mapper\ContentUpdateMapper;
@@ -46,6 +48,8 @@ class ContentEditController extends Controller
 
     /**
      * @var string
+     *
+     * @deprecated Deprecated since 1.10 and will be removed in 2.0. See setPagelayout().
      */
     private $pagelayout;
 
@@ -69,7 +73,7 @@ class ContentEditController extends Controller
      * @param int $parentLocationId Location the content should be a child of
      * @param \Symfony\Component\HttpFoundation\Request $request
      *
-     * @return \Symfony\Component\HttpFoundation\Response
+     * @return \EzSystems\RepositoryForms\Content\View\ContentEditView|\Symfony\Component\HttpFoundation\Response
      */
     public function createWithoutDraftAction($contentTypeIdentifier, $language, $parentLocationId, Request $request)
     {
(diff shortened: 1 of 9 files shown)


bundle/Controller/ContentEditController.php
```diff
@@ -69,7 +73,7 @@ public function __construct(
      * @param int $parentLocationId Location the content should be a child of
      * @param \Symfony\Component\HttpFoundation\Request $request
      *
-     * @return \Symfony\Component\HttpFoundation\Response
+     * @return \EzSystems\RepositoryForms\Content\View\ContentEditView|\Symfony\Component\HttpFoundation\Response
```
alongosz: Note: I'm not a huge fan of returning multiple unrelated object types. Of course refactoring everything now does not make sense, but please be aware of technical debt it creates when in some distant future we'll decide to use strict PHP7 :)

bundle/DependencyInjection/Configuration/Parser/ContentEdit.php
```diff
@@ -0,0 +1,68 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ *
+ * @version //autogentag//
```
alongosz: Nitpick: we no longer use `@version` for anything, to be removed in new code.

bundle/Resources/config/services.yml
```diff
@@ -430,15 +431,21 @@ services:
...
         calls:
             - [setViewTemplate, ['EzSystems\RepositoryForms\UserRegister\View\UserRegisterFormView', "$user_registration.templates.form$"]]
             - [setViewTemplate, ['EzSystems\RepositoryForms\UserRegister\View\UserRegisterConfirmView', "$user_registration.templates.confirmation$"]]
+            - [setViewTemplate, ['EzSystems\RepositoryForms\Content\View\ContentEditView', "$content_edit.templates.edit$"]]
+            - [setViewTemplate, ['EzSystems\RepositoryForms\Content\View\ContentCreateDraftView', "$content_edit.templates.create_draft$"]]
             - [setPagelayout, ["$pagelayout$"]]
 
+    ezrepoforms.user_register.view_templates_listener:
+        alias: ezrepoforms.view_templates_listener
+        depracated: 'The "%service_id%" service is deprecated since 1.10 and will be removed in 2.0. Use "ezrepoforms.view_templates_listener" instead.'
```
alongosz: Typo: `deprecated:`

bundle/Controller/ContentEditController.php
```diff
@@ -88,10 +92,9 @@ public function createWithoutDraftAction($contentTypeIdentifier, $language, $par
             }
         }
 
-        return $this->render('EzSystemsRepositoryFormsBundle:Content:content_edit.html.twig', [
+        return new ContentEditView(null, [
             'form' => $form->createView(),
             'languageCode' => $language,
-            'pagelayout' => $this->pagelayout,
```
alongosz: Is this setting passed to the view somewhere else? I'm not a Twig expert, just asking :)

webhdx: Yes. `pagelayout` is now passed via `ViewTemplatesListener`. This allows us to attach different pagelayout for admin UI.

alongosz: Ah, ok, I see now :)

webhdx:
@alongosz Fixed.
@Nattfarinn Please take a look we need to merge this ASAP.

bundle/Controller/ContentEditController.php
webhdx: Sure. Unfortunately we can't avoid doing so. `Response` objects doesn't allow any extensibility because they come with rendered HTML.

PR #160, opened 2017-10-05
[WIP] Bootstrap 4 support

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-27898

# TODO:
- [ ] Behat tests

### DIFF
--- a/bundle/Resources/views/Content/FieldType/ezauthor.html.twig
+++ b/bundle/Resources/views/Content/FieldType/ezauthor.html.twig
@@ -24,15 +24,23 @@
 {%- endblock ezrepoforms_fieldtype_ezauthor_authors_entry_row -%}
 
 {%- block ezrepoforms_fieldtype_ezauthor_authors_entry_widget -%}
-    <div class="form-row">
+    <div class="form-row rounded ez-author-row">
         {% set row_attr = row_attr|default({})|merge({'class': row_attr.class|default('') ~ ' col'|trim}) %}
         {#{{ block('form_rows') }}#}
         {% for child in form %}
             {{ form_row(child, {'row_attr': row_attr}) }}
         {% endfor %}
         <div class="ez-author-tools col">
-            <button type="button" class="ez-button ez-button-secondary ez-js-remove-author">{{ 'Remove'|trans }}</button>
-            <button type="button" class="ez-button ez-button-secondary ez-js-add-author">{{ 'Add'|trans }}</button>
+            <button type="button" disabled class="ez-button btn btn-icon btn-secondary ez-button-secondary ez-js-remove-author">
+                <svg class="ez-icon ez-icon-trash">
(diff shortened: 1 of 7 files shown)


andrerom:
I guess this would be for v2 only?

alongosz:
> I guess this would be for v2 only?

Yes. Mostly question to @webhdx - is this BC with eZ Platform v1? If not then maybe we should consider releasing `repository-forms` `v2.0` and targeting this to `2.0` branch.

BTW. The bundle linked in the description should be available on `ezsystems` later today.

andrerom:
> Yes. Mostly question to @webhdx - is this BC with eZ Platform v1? If not then maybe we should consider releasing repository-forms v2.0 and targeting this to 2.0 branch.

well if we release or not is not that important until we do official alpha/beta, but 2.0 branch which stays ahead of master somewhat like we do on kernel would be the way to solve it.

webhdx:
@andrerom @alongosz Yes and no. Content editing hasn't been used yet. In v1, repository-forms is only used for content types, policies, roles, languages etc. This PR doesn't change how these forms are rendered.

Technically it's a BC break as it changes HTML markup for content editing, assuming there are people who are already using unfinished content editing. I will create `2.0` branch, we will have more flexibility in changing more stuff to play along with eZ Platform v2 nicely.

webhdx:
@andrerom @alongosz Yes and no. Content editing hasn't been used yet. In v1, repository-forms is only used for content types, policies, roles, languages etc. This PR doesn't change how these forms are rendered.

I'm not sure how we look at it. Changing HTML markup feels like a BC break, i.e. some people might've already styled their forms using our CSS class names. This is still a big stretch as content editing is unfinished and it's hard to assume anyone is using it in the state it is right now. 

On the other hand we already did a BC break merging templating changes for HybridUI in many recent PRs. @andrerom how should we approach another change?

andrerom:
> In v1, repository-forms is only used for content types, policies, roles, languages etc. 

there is also content create, aka UGC, which would benefit from the work on adding more field types.

I'm unsure if html structure can be considered a BC break or not, it would need to be documented yes, but I'll leave it up to @lserwatka to decide if master should simply be aliased to 2.0.x-dev to signify we have BC breaks, or if we can release these changes to 1.12 also.

lserwatka:
@dew326 could you rebase here?

bdunogier:
I'd indeed expect this to be part of a v2. It doesn't cost much, and prevents any backward compatibility issue.

lserwatka:
I confirm, this will be part of v2 @bdunogier, @alongosz

bdunogier:
> I confirm, this will be part of v2 @bdunogier, @alongosz

Perfect, thanks :)

lserwatka:
Closing in favour of https://github.com/ezsystems/repository-forms/pull/163

PR #159, opened 2017-09-25
EZP-24699: Enrich policy limitations view with human readable names

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-24699

## Description

_Disclaimer: There wasn't significant activity in original JIRA ticket since 2016, so I decided to create this  PR as my own open-source contribution_ :wink:

This PR contains implementation of mapping the policy limitation values to human readable versions. 

zrzut ekranu z 2017-09-25 22-43-47

## Architecture

Limitation values are rendered via `ez_render_limitation_value` Twig function, which takes two arguments: 

1. `eZ\Publish\API\Repository\Values\User\Limitation` 
2. Optional hash of parameters

and delegates execution to `EzSystems\RepositoryForms\Templating\LimitationBlockRenderer`. 

Limitation values are mapped by services implementing the following interface: 

```php
/**
 * Interface for Limitation Value mappers.
 */
interface LimitationValueMapperInterface
{
    /**
     * Map the limitation values, in order to pass them as context of limitation value rendering.
     *
     * @param Limitation $limitation
     * @return mixed[]
     */
    public function mapLimitationValue(Limitation $limitation);
}
```
and registered in DIC with `ez.limitation.valueMapper` tag with `limitationType` attribute e.g.

```yml
ezrepoforms.limitation.form_mapper.newsection:
    parent: ezrepoforms.limitation.form_mapper.multiple_selection
    class: "%ezrepoforms.limitation.form_mapper.section.class%"
    arguments: ["@ezpublish.api.service.section"]
    tags:
        - { name: ez.limitation.formMapper, limitationType: NewSection }
        - { name: ez.limitation.valueMapper, limitationType: NewSection }
```

Responsibility of `EzSystems\RepositoryForms\Templating\LimitationRenderer` is similar to `\eZ\Publish\Core\MVC\Symfony\Templating\Twig\FieldBlockRenderer`: render propery twig block and supply all required parameters: 

* limitation instance
* mapped values 
* parameters (second argument of `ez_render_limitation_value`) 

Block name is resolved as `ez_limitation_%LIMITATION_IDENTIFIER%_value`
, where `%LIMITATION_IDENTIFIER%` is identifier of limitation (converted to lower case characters). 

If the value mapper for limitation type is not provided, the "fallback block" is rendered (`ez_limitation_value_fallback`).

## TODO:

- [X] Implementation
- [x] Tests
[...]

PR #162, opened 2017-10-16
EZP-28077: As a bundle developer, I want to easily provide template for custom limitation value

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28077

# Description

Improvement for https://github.com/ezsystems/repository-forms/pull/159: Overriding parameters defined by 3rd party bundle is not easy for bundle developer. This PR adds an ability to configure the list of resources containing definitions of limitation value blocks, in similar to fields view/edit/definition view/definition edit blocks. Configuration example: 

```yaml
system:
    default:
        limitation_value_templates:
            - { template: 'AppBundle:Limitation:custom_limitation_value.html.twig', priority: 0 }
```

### DIFF
--- a/bundle/DependencyInjection/Configuration/Parser/LimitationValueTemplates.php
+++ b/bundle/DependencyInjection/Configuration/Parser/LimitationValueTemplates.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryFormsBundle\DependencyInjection\Configuration\Parser;
+
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser\Templates;
+
+class LimitationValueTemplates extends Templates
+{
+    const NODE_KEY = 'limitation_value_templates';
+    const INFO = 'Settings for limitation value templates';
+    const INFO_TEMPLATE_KEY = 'Template file where to find block definition to display limitation values';
+}
--- a/lib/Limitation/Templating/LimitationBlockRenderer.php
+++ b/lib/Limitation/Templating/LimitationBlockRenderer.php
@@ -70,9 +70,13 @@ class LimitationBlockRenderer implements LimitationBlockRendererInterface
         return $template->renderBlock($blockName, $parameters);
     }
 
-    public function setLimitationValueResources(array $limitationValueResources)
+    public function setLimitationValueResources(array $resources)
     {
-        $this->limitationValueResources = $limitationValueResources;
+        usort($resources, function ($a, $b) {
+            return $b['priority'] - $a['priority'];
+        });
+
+        $this->limitationValueResources = array_column($resources, 'template');
     }
 
     /**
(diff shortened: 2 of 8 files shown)


andrerom:
@adamwojs this should aim for 1.10 branch _(and once this is in we should tag a beta of this @lserwatka)_.

adamwojs:
@andrerom Thanks, target branch has been changed.

bundle/DependencyInjection/Configuration/Parser/LimitationValueTemplates.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
```
alongosz: Please add license info PhpDoc

tests/RepositoryForms/Twig/LimitationValueRenderingExtensionTest.php
```diff
@@ -32,8 +32,14 @@ public function getExtensions(\Twig_Environment $twig = null)
...
-            'templates/limitation_value_1.html.twig',
-            'templates/limitation_value_2.html.twig',
+            [
+                'template' => 'templates/limitation_value_1.html.twig',
+                'priority' => 0,
+            ],
+            [
+                'templates/limitation_value_2.html.twig',
+                'priority' => 0,
+            ],
```
alongosz: Since the feature includes parsing of priorities, can you set different priorities here and test them? (unless it's covered elsewhere).

bundle/DependencyInjection/Configuration/Parser/LimitationValueTemplates.php
Nattfarinn: Just run `php-cs-fixer`. It should add the missing header.

lib/Limitation/Templating/LimitationBlockRenderer.php
```diff
@@ -70,9 +70,13 @@ public function renderLimitationValue(Limitation $limitation, array $parameters
         return $template->renderBlock($blockName, $parameters);
     }
 
-    public function setLimitationValueResources(array $limitationValueResources)
+    public function setLimitationValueResources(array $resources)
     {
-        $this->limitationValueResources = $limitationValueResources;
+        usort($resources, function ($a, $b) {
+            return $b['priority'] - $a['priority'];
```
webhdx: Minor thing but since we encourage to use PHP 7 you could have used spaceship operator here (<=>).

alongosz:
> Just run php-cs-fixer. It should add the missing header.

@Nattfarinn Actually I don't see header config for this repository. Some time ago we decided not to add this for existing and released packages as backporting could cause conflicts. We can revisit this decision of course.

*(That's why I pushed to add this to our new admin package until it's too late btw ;) )*

lib/Limitation/Templating/LimitationBlockRenderer.php
alongosz: > Minor thing but since we encourage to use PHP 7 you could have used spaceship operator here (<=>).

Noooo... :) It's `1.x`, so should work with PHP 5.6

bdunogier:
One general thought about this, that might apply to other packages: should we consider using `ezplatform-design` to refer to templates ?

With it, we could make our templates part of `standard`, and when applicable `admin`, and give full flexibility to integrators to override any of the templates by overriding them in their own design(s).

adamwojs:
PR updated according to suggestions.

lserwatka:
@bdunogier ezplatform-design is something we will add in the next version. We can't handle it for v2 due to limited capacity.

andrerom:
@bdunogier If `ezplatform-design` is changed as discussed to do rather fallback to symfony default placement for files, then any bundle, including this one will effectively become themable. That said there are [changes to overriding templates in symfony 3.4 / 4.0](https://symfony.com/blog/new-in-symfony-3-4-improved-the-overriding-of-templates) maybe worth checking out over the next month, might or might not simplify things like the [new builtin lazy loaded auto inecjetion of tagged services feature](https://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services) will.

PR #163, opened 2017-10-16
EZP-28038: As a Developer I would like to have Symfony form Types representing every eZ Platform fieldtype

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28038

# Description
This PR provides better code organization for field types integration. The idea was to have every field type defined as a separate Symfony FormType which gives more flexibility for extending functionality and theming.

Some fieldtypes (i.e. `ezemail`) that doesn't come with definition settings are defined solely in `services.yml` using `FormTypeBasedFieldValueFormMapper` and I decided to leave them as-is. It would be appropriate if they need some special theming but this isn't the case.

# TODO
- [x] fix unit tests
- [x] fix behat tests
- [x] document all BC breaks (actually there are many of them)
- [x] move all fieldtypes related service definitions to `fieldtypes.yml`


webhdx:
@alongosz @Nattfarinn @mikadamczyk Rebased and reorganized commit for easier reviewing.

PR #161, opened 2017-10-13
[CS] Aligned php-cs-fixer config with changes introduced by v2.7.1

alongosz:
This PR updates php-cs-fixer config to align with changes introduced by `v2.7.1`.

**Backporting this** to `1.5` as this is the oldest branch supported by `1.7 LTS`.

**TODO**:
- [x] Add `'yoda_style' => false` setting
- [x] Add `'no_break_comment' => false` setting

PR #165, opened 2017-10-17
Added `friendsofphp/php-cs-fixer` requirement to `composer.json`

webhdx:
# Description
#161 introduced minimal requirement on `php-cs-fixer` 2.7 but it wasn't specified in the `composer.json`. To keep things clear I added this so that developers can fix their code using bundled (`bin/php-cs-fixer`) fixer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,9 @@
             "dev-master": "1.5.x-dev"
         }
     },
+    "scripts": {
+        "fix": "@php bin/php-cs-fixer fix"
+    },
     "config": {
         "bin-dir": "bin"
     }
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
     "require-dev": {
         "phpunit/phpunit": "^5.7",
         "matthiasnoback/symfony-dependency-injection-test": "~1.0",
-        "behat/behat": "^3.0"
+        "behat/behat": "^3.0",
+        "friendsofphp/php-cs-fixer": "^2.7"
     },
     "conflict": {
         "symfony/symfony": "2.7.11"


andrerom:
hmm, it's typically installed globally, but of course if we intend to do this in all repos and change ezrobot to use the version repo supports to avoid all the breaks everytime @vidarl is forced to upgrade it I'm all for it.

webhdx:
Indeed but I had version 2.1 installed on my machine and it turned out it's incompatible with the configuration provided in `.php_cs`. If we don't provide minimal version requirement then we are forcing developers to guess which version is needed. @ezrobot is one thing but keep in mind that `php-cs-fixer` is also used by people who are proposing PR's.

MarioBlazek:
@webhdx it would be nice to also have something like [this](https://github.com/netgen/NetgenInformationCollectionBundle/blob/master/composer.json#L47).

webhdx:
@MarioBlazek Very good idea. I've added script definition.

alongosz:
> hmm, it's typically installed globally, but of course if we intend to do this in all repos and change ezrobot to use the version repo supports to avoid all the breaks everytime @vidarl is forced to upgrade it I'm all for it.

I was even thinking about integrating ezrobot's work into Travis jobs (didn't we intend to move everything to Travis anyway?)

Pros:
- No more global version update issues.
- Full control over coding standards per repository.
- No need to bother Vidar with ezrobot integration.

Cons:
- no diffs in comments (this is annoying and useless IMHO, but maybe someone uses it?), actual diff would be accessible in Travis output.
- a need for a new script doing this check (I can write one, attempting to make it generic for every repo).
- more work for already overworked Travis.

// cc @andrerom @lserwatka 

> it would be nice to also have something like [this](https://github.com/netgen/NetgenInformationCollectionBundle/blob/master/composer.json#L47).

That's indeed a good idea :+1:

lserwatka:
@webhdx could you please merge this up?

MarioBlazek:
@alongosz you can use CodeClimate for those syntax checks

alongosz:
> @alongosz you can use CodeClimate for those syntax checks

Lot's of possibilities, what I try to achieve is to keep everything in one place :)

MarioBlazek:
@alongosz but why ? Travis CI is not meant to do everything :)

alongosz:
Less time to configure CI for new repositories. Less people to ping :) 
But if some other CI tool is more suitable for some tasks, then I would accept it and use it.

MarioBlazek:
I would suggest using TravisCI for running tests, and CodeClimate for code related stuff (checking CS, etc).

alongosz:
> I would suggest using TravisCI for running tests, and CodeClimate for code related stuff (checking CS, etc).

Sounds like a way to consider :+1:

PR #168, opened 2017-10-27
Update Travis config for 1.10 branch

mnocon:
I've noticed that Travis does not list 1.10 as active branch:
https://travis-ci.org/ezsystems/repository-forms/branches
and I think it should (and unless I'm wrong this should help?).

I've used GitHub online editor - please let me know if I should create a JIRA ticket (and have a properly named branch), I'll adjust.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
 branches:
   only:
     - master
-    - "1.9"
+    - "1.10"
     - "/^feature-/"
 
 before_install:


mnocon:
Hi @alongosz, @andrerom  - could you have a look if this change makes sense? I remember something like this has been done for 1.5 branch: 
https://github.com/ezsystems/repository-forms/commit/6190e28fc5692bd502cdbec8f1218787a73623a7

.travis.yml
```diff
@@ -21,7 +21,7 @@ matrix:
 branches:
   only:
     - master
-    - "1.9"
```
alongosz: Since `1.10` is still `rc`, IMHO we should support `1.9`. POV @lserwatka @andrerom?

lserwatka: Why? It it should point to 1.10 branch.

andrerom: This is 1.10 branch, so it should be 1.10, otherwise 1.10 will never be built.


1.9 will still be built ;)

alongosz: > This is 1.10 branch (...)
> 1.9 will still be built ;)

Ahm... yes, good point :)

PR #169, opened 2017-11-09
EZP-28204: Change view data format to timestamp for TimeFieldType

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28204

# Description
This PR changes `eztime` form type data format to a timestamp which is easier to be handled by a JS datepicker components.

### DIFF
--- a/lib/FieldType/DataTransformer/TimeValueTransformer.php
+++ b/lib/FieldType/DataTransformer/TimeValueTransformer.php
@@ -1,0 +1,65 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\FieldType\DataTransformer;
+
+use eZ\Publish\API\Repository\Exceptions\InvalidArgumentException;
+use eZ\Publish\Core\FieldType\Time\Value;
+use Symfony\Component\Form\DataTransformerInterface;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+
+/**
+ * DataTransformer for Time\Value.
+ */
+class TimeValueTransformer implements DataTransformerInterface
+{
+    /**
+     * @param Value|mixed $value
+     *
+     * @return int|null
+     *
+     * @throws TransformationFailedException
+     */
+    public function transform($value)
+    {
+        if (null === $value->time) {
+            return null;
+        }
+
+        if (!$value instanceof Value) {
+            throw new TransformationFailedException(
+                sprintf('Expected a %s', Value::class)
+            );
+        }
+
+        return $value->time;
+    }
+
+    /**
+     * @param int|mixed $value
+     *
+     * @return Value|null
+     *
+     * @throws InvalidArgumentException
+     * @throws TransformationFailedException
+     */
+    public function reverseTransform($value)
+    {
+        if (empty($value)) {
+            return null;
+        }
+
+        if (!is_numeric($value)) {
+            throw new TransformationFailedException(
+                sprintf('Expected a numeric, got %s instead', gettype($value))
+            );
+        }
+
+        return Value::fromTimestamp($value);
+    }
+}
--- a/lib/Form/Type/FieldType/TimeFieldType.php
+++ b/lib/Form/Type/FieldType/TimeFieldType.php
@@ -5,10 +5,9 @@
  */
 namespace EzSystems\RepositoryForms\Form\Type\FieldType;
 
-use eZ\Publish\API\Repository\FieldTypeService;
-use EzSystems\RepositoryForms\FieldType\DataTransformer\FieldValueTransformer;
+use EzSystems\RepositoryForms\FieldType\DataTransformer\TimeValueTransformer;
 use Symfony\Component\Form\AbstractType;
-use Symfony\Component\Form\Extension\Core\Type\TimeType;
+use Symfony\Component\Form\Extension\Core\Type\IntegerType;
 use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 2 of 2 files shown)


lib/FieldType/DataTransformer/FormattedTimeToTimestampTransformer.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
```
alongosz: Nitpick: please skip

lib/FieldType/DataTransformer/TimeValueTransformer.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
```
alongosz: Nitpick: skip

lib/Form/Type/FieldType/TimeFieldType.php
```diff
@@ -44,21 +45,21 @@ public function getParent()
 
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->addModelTransformer(new FieldValueTransformer($this->fieldTypeService->getFieldType('eztime')));
+        $builder
+            ->addViewTransformer(new FormattedTimeToTimestampTransformer($options['with_seconds']))
```
alongosz: Are you sure `with_seconds` key will be always present?

lib/Form/Type/FieldType/TimeFieldType.php
```diff
@@ -44,21 +45,21 @@ public function getParent()
...
+            ->addModelTransformer(new TimeValueTransformer());
     }
 
     public function buildView(FormView $view, FormInterface $form, array $options)
     {
-        if ($options['with_seconds']) {
-            $view->vars['attr']['step'] = 1;
-        }
+        $view->vars['attr']['data-seconds'] = (int) $options['with_seconds'];
```
alongosz: Ok, so is this option a `bool` (inferring from `FormattedTimeToTimestampTransformer` constructor call) or an `int`?
The "thing" on the other side, using view variables should be able to handle bool.
But I don't know if it's possible for forms, if not then this is fine.

lib/Form/Type/FieldType/TimeFieldType.php
```diff
@@ -44,21 +45,21 @@ public function getParent()
 
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->addModelTransformer(new FieldValueTransformer($this->fieldTypeService->getFieldType('eztime')));
+        $builder
+            ->addViewTransformer(new FormattedTimeToTimestampTransformer($options['with_seconds']))
+            ->addModelTransformer(new TimeValueTransformer());
```
alongosz: Noob question: what's the difference between Model and View transformers in this case? Based on [Symfony Doc](http://symfony.com/doc/current/form/data_transformers.html#about-model-and-view-transformers) I'd expect both transformers to be Model transformers, but I don't know this that well (or rather at all :stuck_out_tongue_winking_eye:)

lib/FieldType/DataTransformer/FormattedTimeToTimestampTransformer.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        if (empty($value)) {
+            return null;
+        }
+
+        if (!is_string($value)) {
+            throw new TransformationFailedException('Expected a string.');
+        }
+
+        $dateTime = DateTime::createFromFormat($this->withSeconds ? 'H:i:s' : 'H:i', $value);
```
mikadamczyk: Is it possible that the function `createFromFormat()` will return bool in this case?

lib/Form/Type/FieldType/TimeFieldType.php
webhdx: There are 3 types of data in Symfony forms:
* Model data - used by your model
* Normalized - used by the form type itself
* View data - used as a value in HTML

Model Transformer is doing the conversion between Model<->Norm data.
View Transformer is doing the conversion between Norm<->View data.

lib/Form/Type/FieldType/TimeFieldType.php
webhdx: It a `bool` value but we need to present is as either `0` or `1` as `data-*` HTML attributes do not support boolean values.

lib/Form/Type/FieldType/TimeFieldType.php
webhdx: Yes. It has default value always set.

lib/FieldType/DataTransformer/FormattedTimeToTimestampTransformer.php
webhdx: I don't think so? Why would it return `true`? It should return a proper object or `false` (gotta love PHP for that inconsistency) if the `$value` cannot be parsed using provided datetime format.

mikadamczyk: If it will return `false` for some reason, then next line will fail

MarioBlazek: This line will fail if you provide timestamp in `$value`.

webhdx:
@Nattfarinn @alongosz @mikadamczyk Please note I changed my implementation. I over engineered it previously. There was no need to inherit `TimeType` and force it to use `DateTime` as a model data format (it was kinda hackish). Simple `IntegerType` will do the job. Sorry for the inconvenience.

PR #177, opened 2017-11-21
EZP-28287: Inconsistent eztime value across timezones

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28287

# Description
This PR fixes data transformation for `eztime` fieldtype. Previously it was using timestamps which wasn't really working - set time should be timezone independent. Right now value represents seconds which follow business logic behind `eztime`.


MarioBlazek:
Nice.

andrerom:
datetime affected by this?

webhdx:
@andrerom Nope, `ezdatetime` is fine.

PR #178, opened 2017-11-22
EZP-28299: Fixes default settings value initialization for binary based fieldtypes

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28299

# Description
This PR fixes the problem with empty settings like width and height (in case of `ezmedia` field type). Type expects an integer for those thus it was causing an error. Data Transformers are now using default values from an empty field value.


webhdx:
@alongosz No problem, changed to `getDefaultProperties()`.

PR #179, opened 2017-11-24
EZP-28315: Cannot delete draft in case of nonpublished content

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28315
> Requires: https://github.com/ezsystems/ezpublish-kernel/pull/2163

# Description
This PR fixes the problem of deleting a draft when there is only one version in edited content. Kernel doesn't allow to remove version if that's the only one in the content. In order to fix that it's required to detect single version content and remove content completely instead. The fix relies on kernel improvement: https://github.com/ezsystems/ezpublish-kernel/pull/2163

**TODO**:
- [x] Wait for ezpublish-kernel/#2163 getting merged/rebased into `7.0`

### DIFF
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -193,8 +193,11 @@ services:
             - { name: kernel.event_subscriber }
 
     ezrepoforms.form_processor.content:
-        class: "%ezrepoforms.form_processor.content.class%"
-        arguments: ["@ezpublish.api.service.content", "@router"]
+        class: '%ezrepoforms.form_processor.content.class%'
+        arguments:
+            - '@ezpublish.api.service.content'
+            - '@ezpublish.api.service.location'
+            - '@router'
         tags:
             - { name: kernel.event_subscriber }
 
--- a/lib/Form/Processor/ContentFormProcessor.php
+++ b/lib/Form/Processor/ContentFormProcessor.php
@@ -9,6 +9,7 @@
 namespace EzSystems\RepositoryForms\Form\Processor;
 
 use eZ\Publish\API\Repository\ContentService;
+use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\ContentStruct;
 use eZ\Publish\Core\MVC\Symfony\Routing\UrlAliasRouter;
 use EzSystems\RepositoryForms\Event\FormActionEvent;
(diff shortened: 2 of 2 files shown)


bundle/Resources/config/services.yml
```diff
@@ -194,7 +194,10 @@ services:
 
     ezrepoforms.form_processor.content:
         class: "%ezrepoforms.form_processor.content.class%"
-        arguments: ["@ezpublish.api.service.content", "@router"]
+        arguments:
+            - "@ezpublish.api.service.content"
+            - "@ezpublish.api.service.location"
+            - "@router"
```
alongosz: Hmm, how about we gradually change double to single quotes (meaning for service being updated only)?
*I know this makes file inconsistent, but it's the only way to get rid of old convention without making "diff mess".*

lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -100,10 +104,23 @@ public function processCancel(FormActionEvent $event)
...
 
-        $this->contentService->deleteVersion($data->contentDraft->getVersionInfo());
+        $content = $data->contentDraft;
+        $contentInfo = $content->contentInfo;
+        $versionInfo = $data->contentDraft->getVersionInfo();
+
+        // if there is only one version you have to remove whole content instead of a version itself
+        if (1 === count($this->contentService->loadVersions($contentInfo))) {
+            $location = $this->locationService->loadParentLocationsForDraftContent($versionInfo)[0];
+            $contentId = $location->getContentInfo()->id; // parent location content id
```
alongosz: Hmm, if this is the only point when you needed Location, you actually didn't need it at all. `$contentId` can be obtained from `$contentInfo` which is above this `if`.

webhdx: This is a `$contentId` of parent location this is why I can't use `$contentInfo`. Maybe variable names are a little bit misleading, I will rename these.

alongosz: Ah, right. I need normal :coffee:, not the imitation I drank this morning...

webhdx:
@alongosz Fixed.

lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -100,10 +104,23 @@ public function processCancel(FormActionEvent $event)
             return;
         }
 
-        $this->contentService->deleteVersion($data->contentDraft->getVersionInfo());
+        $content = $data->contentDraft;
+        $contentInfo = $content->contentInfo;
+        $versionInfo = $data->contentDraft->getVersionInfo();
+
+        // if there is only one version you have to remove whole content instead of a version itself
+        if (1 === count($this->contentService->loadVersions($contentInfo))) {
```
andrerom: API DX improvement potential here /cc @alongosz @bdunogier   :)

PR #180, opened 2017-11-30
EZP-28342: Invalid input type for latitude and logitude in MapLocation fieldtype

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28342

# Description
MapLocation's children types had invalid input type (text instead of number) which made validation fail and also it was incorrect from HTML5 standpoint as text input should not have `min` and `max` attributes.

### DIFF
--- a/lib/Form/Type/FieldType/MapLocationFieldType.php
+++ b/lib/Form/Type/FieldType/MapLocationFieldType.php
@@ -11,6 +11,9 @@ use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\NumberType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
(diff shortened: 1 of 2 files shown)


lib/Form/Type/FieldType/MapLocationFieldType.php
```diff
@@ -42,36 +44,41 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
                 NumberType::class,
                 [
-                    'label' => 'content.field_type.ezgmaplocation.latitude',
+                    'label' => /** @Desc("Latitude") */ 'content.field_type.ezgmaplocation.latitude',
                     'required' => $options['required'],
                     'attr' => [
                         'min' => -90,
                         'max' => 90,
-                        'step' => 'any',
+                        'step' => 0.000001,
```
alongosz: Is this option required at all? I don't see a point in defining steps for lat/long numbers as they're very precise and changing them by decreasing/increasing has no value.

webhdx: We want to use native HTML5 inputs, in this case it's `number` input type. User is still allowed to provide value manually.

alongosz: If it's required from HTML5 standpoint, then no objections.

PR #182, opened 2017-12-01
EZP-27539: Translation support in content edit

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-27539

# Description

Translation support was never implemented in repository-forms. Now there is a need for that in new AdminUI. This PR makes fieldtypes "language" aware which is helpful for developers theming content edit page. It also fixes the issue of nontranslated labels of fieldtypes as this feature is there but never implemented in Admin UI or Platform UI.

### DIFF
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -11,6 +11,7 @@ namespace EzSystems\RepositoryFormsBundle\Controller;
 use eZ\Bundle\EzPublishCoreBundle\Controller;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\LanguageService;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo;
 use eZ\Publish\Core\Base\Exceptions\BadStateException;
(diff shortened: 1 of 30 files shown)


bundle/Controller/ContentEditController.php
```diff
@@ -103,7 +103,7 @@ public function createWithoutDraftAction($contentTypeIdentifier, $language, $par
 
         return new ContentCreateView(null, [
             'form' => $form->createView(),
-            'languageCode' => $language,
```
bdunogier: It might not be critical, but maybe I'd try to keep some BC here: _add_ `language`, and comment `languageCode` as deprecated ?

lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -153,7 +156,12 @@ public function processCreateDraft(FormActionEvent $event)
      */
     private function saveDraft(ContentStruct $data, $languageCode)
     {
+        $mainLanuageCode = $this->resolveMainLanguageCode($data);
```
bdunogier: Typo on `$mainLanguageCode`.

lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -165,4 +173,16 @@ private function saveDraft(ContentStruct $data, $languageCode)
...
     }
+
+    /**
+     * @param ContentUpdateData|ContentCreateData|NewnessChecker $data
+     *
+     * @return string
+     */
+    private function resolveMainLanguageCode($data): string
+    {
+        return $data->isNew()
```
bdunogier: Should we be a tiny bit protective and test the type of `$data`, since we don't have typehinting here ?

webhdx: Good idea. Fixed.

lib/Form/Processor/ContentFormProcessor.php
webhdx: Argh, happens to me on word *language* constantly ;)

bundle/Controller/ContentEditController.php
webhdx: I'm not sure. 2.0 will change a lot actually, I'd love to make it as clean as possible in 2.0 so that we aren't left with a lot of deprecations till 3.0. I believe it's just a matter of writing a descriptive doc on what is not backwards compatible.

PR #191, opened 2017-12-14
EZP-28531: Force calling FieldDefinition->getName() with mainLanguageCode

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28531


andrerom:
@lserwatka I'll leave this up to you, from the looks if it the code now forces main language if it is present, otherwise fallback to first translated name in the list which I don't think is ordered by priority.  So from the looks of this, it seems to break the language priority logic to solve issues in tests.

But I might be wrong, I'm definitely not fluent in repo-forms logic.

PR #192, opened 2017-12-14
EZP-28532: Fix User related tests

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28532


webhdx:
@andrerom Yes, we need to merge all 3 pull requests. I didn't want to make one PR because some things are needed from 1.5 and up.

PR #194, opened 2017-12-14
[Travis] Use Trusty

andrerom:
- The PHP 5.6 failure is still the case here even so there seems to be a remaining issue here, for some reason on PHP 5.6 /cc @webhdx 
- [x] Behat is failing here so code here needs to use newer version of the setups script from meta (the trusty variant)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
 branches:
   only:
     - master
-    - "1.5"
+    - /^\d.\d+$/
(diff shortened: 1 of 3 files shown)


andrerom:
Merging, makes behat at least run, and now failing on a few broken tests. I'll leave it to @ezsystems/ez-qa-team / dev team to look into what needs to be fixed for that.

unit tests on 5.6 still broken for unknown reason.

PR #211, opened 2018-01-16
Enable PHPUnit 6

MarioBlazek:


andrerom:
Behat failure afaik not related.

webhdx:
Yeah I need to take a look at behat tests. PR is alright.

PR #214, opened 2018-02-16
EZP-28850: Add optional location parameter to draft edit for more appropriate redirection

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28850
> Requires: https://github.com/ezsystems/ezplatform-admin-ui/pull/374

# Description
Right now it's not possible to return to the proper location in AdminUI as repository-forms is not passing any information about the location draft was edited from. It's also crucial for Page Builder. 

The parameter is optional and should not introduce any BC breaks.

# QA Testing
I'm not sure how are you supposed to test it, the best thing that comes to my mind is to just test content edit in Admin UI as this is directly affected.

### DIFF
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -24,6 +24,7 @@ use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\Content\ContentDraftCreateType;
 use EzSystems\RepositoryForms\Form\Type\Content\ContentEditType;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 
 class ContentEditController extends Controller
 {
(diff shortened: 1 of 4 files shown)


webhdx:
Weird but I've checked and it looks like HTTP spec is using misspelled "referer" (it was my inspiration). I've changed it though. Thanks @bdunogier.

barbaragr:
1. Create new Article, fill required fields
2. Click Publish
3. Click Edit, change nothing
4. Click Publish or Save  - there is an error 500:

> The option "referrerLocation" with value null is expected to be of type "eZ\Publish\API\Repository\Values\Content\Location", but is of type "NULL".

lserwatka:
@webhdx did you address issue from @barbaragr ?

webhdx:
Not yet @lserwatka I had a few other things going on. I'm on it right now tho.

webhdx:
@barbaragr Fixed but please note you are now required to also use https://github.com/ezsystems/ezplatform-admin-ui/pull/374

barbaragr:
When I click "Browse" nothing happen and console shows: 
> Uncaught TypeError: tbody.rows.count is not a function

PR #215, opened 2018-02-16
EZP-28849: Allow repository forms to use View Manager for handling draft edit controller and template dispatching

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28849
> Requires: https://github.com/ezsystems/ezplatform-admin-ui/pull/375 (to work with AdminUI)

# Description
I heard from @bdunogier it was always the initial idea but no one had time to finish it. As Page Builder is meant to replace edit interface for Landing Pages we need this feature. This PR only affects draft edit action, next one will be dedicated to content create.

I'd like a solid code review as this PR might potentially introduce some BC breaks.

# QA Testing
Might be hard to test by QA as this is mostly related to developer experience but I'd advise to get through content editing briefly. If the content editing screen is displayed correctly and the actions are performed as expected it should be alright.

### DIFF
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -12,19 +12,15 @@ use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LanguageService;
 use eZ\Publish\API\Repository\LocationService;
-use eZ\Publish\API\Repository\Values\Content\VersionInfo;
-use eZ\Publish\Core\Base\Exceptions\BadStateException;
 use EzSystems\RepositoryForms\Content\View\ContentCreateDraftView;
 use EzSystems\RepositoryForms\Content\View\ContentCreateView;
 use EzSystems\RepositoryForms\Content\View\ContentEditView;
 use EzSystems\RepositoryForms\Data\Content\CreateContentDraftData;
 use EzSystems\RepositoryForms\Data\Mapper\ContentCreateMapper;
-use EzSystems\RepositoryForms\Data\Mapper\ContentUpdateMapper;
 use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\Content\ContentDraftCreateType;
 use EzSystems\RepositoryForms\Form\Type\Content\ContentEditType;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\Response;
 
 class ContentEditController extends Controller
 {
@@ -42,13 +38,6 @@ class ContentEditController extends Controller
 
     /** @var ActionDispatcherInterface */
     private $contentActionDispatcher;
-
-    /**
-     * @var string
-     *
-     * @deprecated Deprecated since 1.10 and will be removed in 2.0. See setPagelayout().
-     */
-    private $pagelayout;
 
     public function __construct(
         ContentTypeService $contentTypeService,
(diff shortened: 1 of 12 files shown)


lib/EventListener/ContentEditViewParametersInjectorListener.php
```diff
@@ -0,0 +1,42 @@
...
+        return [
+            ViewEvents::FILTER_VIEW_PARAMETERS => ['filterViewParameters', 10],
+        ];
+    }
+
+    public function filterViewParameters(FilterViewParametersEvent $event)
+    {
+        /** @var ContentEditView $view */
+        $view = $event->getView();
+        if (!$view instanceof ContentEditView) {
```
bdunogier: Should `ContentEditView` inherit from `ContentView` ? That way, you wouldn't need to reimplement tis listener. You can of course have an extra one that does edit specific things.

webhdx:
ping @Nattfarinn @mikadamczyk @ViniTou any chance for code review?

bundle/Controller/ContentEditController.php
```diff
@@ -117,7 +110,7 @@ public function createWithoutDraftAction($contentTypeIdentifier, $language, $par
      * @param string $toLanguage
      * @param \Symfony\Component\HttpFoundation\Request $request
      *
-     * @return \EzSystems\RepositoryForms\Content\View\ContentCreateDraftView|\Symfony\Component\HttpFoundation\Response
+     * @return \EzSystems\RepositoryForms\Content\View\ContentCreateDraftView|Response
```
alongosz: There's a lot of inconsistency in PhpDoc type hinting. Please decide to use either FQCN or short name.

bdunogier: To my knowledge, we use fully-qualified namespaces.

While real IDEs will help navigate with short names, you don't get that option when reviewing code on github. I think it's a very valid argument.

webhdx: Depends on the repository. AdminUI uses short names. Is there a way to enforce it with PHP CS Fixer? @alongosz do you know?

alongosz: > Is there a way to enforce it with PHP CS Fixer?

No, I already looked for it some time ago. Checked just now - nothing changed - we probably would need to contribute... :)

webhdx: Ok. I've changed to FQCN.

webhdx:
@mikadamczyk @ViniTou @alongosz @Nattfarinn Code review needed.

webhdx:
@bdunogier Controller won't stay empty because in `ViewBuilder` context you don't have an access to the `Request` thus you can't handle the form. I can get master request from the Container but the question is - is it the proper way to do it? Looks too hardcore for my liking. I'd rather have minimal logic in the controller so that anyone can override default behavior. The View is only passing information which is needed to render the template and it's fine for me. What do you think?

bdunogier:
@webhdx actually, you do get the Request with ViewBuilder Filters. It is what I had used when I had prototyped this: https://github.com/ezsystems/repository-forms/pull/62/files#diff-03ff5c01c6c75308bdf76dbeb0596a18.

lib/Content/View/Builder/ContentEditViewBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+
+    /**
+     * @param array $parameters
+     *
+     * @return \eZ\Publish\Core\MVC\Symfony\View\ContentView|\eZ\Publish\Core\MVC\Symfony\View\View
+     *
+     * @throws \Symfony\Component\OptionsResolver\Exception\InvalidOptionsException
+     * @throws \eZ\Publish\Core\Base\Exceptions\BadStateException
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\UnauthorizedException
```
alongosz: @webhdx please type-hint API

lib/Content/View/Builder/ContentEditViewBuilder.php
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /**
+     * @param array $parameters
+     * @param \eZ\Publish\API\Repository\Values\Content\Location|null $location
+     * @param \eZ\Publish\API\Repository\Values\Content\Language $language
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Content
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentException
```
alongosz: @webhdx please type-hint API

lib/Content/View/Builder/ContentEditViewBuilder.php
```diff
@@ -0,0 +1,298 @@
+<?php
```
alongosz: Nitpick: missing PhpDoc with license and copyright

PR #218, opened 2018-03-05
EZP-28871: upload_max_filesize setting should be taken into account in the validation of uploaded files

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28871

## Description 

The `upload_max_filesize` setting from `php.ini` should be taken into account in the validation of uploaded files, as an upper limit of upload size. 

## Steps to reproduce 

1. Configure `upload_max_filesize` to e.g 5M
1. Create new Media -> File
2. Select large binary file larger then `upload_max_filesize`
3. Try to save content. The application will return "Internal server error" because of  `Argument 'localFile' is invalid: file does not exist or is unreadable: /var/www/...` exception

### DIFF
--- a/lib/Form/Type/FieldType/BinaryBaseFieldType.php
+++ b/lib/Form/Type/FieldType/BinaryBaseFieldType.php
@@ -37,9 +37,9 @@ class BinaryBaseFieldType extends AbstractType
                     'required' => $options['required'],
                     'constraints' => [
                         new Assert\File([
-                            'maxSize' => $this->getMaxUploadSize()
-                        ])
-                    ]
+                            'maxSize' => $this->getMaxUploadSize(),
+                        ]),
+                    ],
                 ]
             );
     }
@@ -69,8 +69,8 @@ class BinaryBaseFieldType extends AbstractType
         $str = strtoupper(trim($str));
 
         $value = substr($str, 0, -1);
-        $unit  = substr($str, -1);
-        switch($unit) {
+        $unit = substr($str, -1);
+        switch ($unit) {
             case 'G':
                 $value *= 1024;
             case 'M':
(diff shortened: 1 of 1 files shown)


crevillo:
What about the ezimage field? Is covered by this?

adamwojs:
@crevillo Yes, it's base class form ezbinary, ezimage and ezmedia

lib/Form/Type/FieldType/BinaryBaseFieldType.php
```diff
@@ -32,12 +35,50 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+
+    private function str2bytes($str)
+    {
+        $str = strtoupper(trim($str));
+
+        $value = substr($str, 0, -1);
+        $unit = substr($str, -1);
+        switch ($unit) {
+            case 'G':
+                $value *= 1024;
```
crevillo: no break here?

adamwojs: Yes, it's intentionally. To convert gigabytes to bytes we want to multiply a value by 1024^3, for M by 1024^2 and so on.  Ref: http://php.net/manual/en/faq.using.php#faq.using.shorthandbytes

crevillo: sorry, then i don't understand. if i put `1G` here will value be 1 * 1024 * 1024 * 1024?

wouldn't it be only `1 * 1024` as `case M` and `case K` will be false?

andrerom: > if i put 1G here will value be 1 * 1024 * 1024 * 1024?

yes

> wouldn't it be only 1 * 1024 as case M and case K will be false?

as you said above, there is no break, so it does not matter if case is false.

So this is why units are sorted by size here, so falltrue is taken advantage of.

crevillo: ok, understood now.

lserwatka:
@adamwojs I think you must change a traget branch to 2.0 if corresponding PR https://github.com/ezsystems/ezplatform-admin-ui/pull/368 targets 2.0.x

PR #221, opened 2018-03-15
EZP-28551: [Media and Binary] You can set maximum file size to less than zero

sunpietro:
https://jira.ez.no/browse/EZP-28551

### DIFF
--- a/bundle/Resources/config/form_types.yml
+++ b/bundle/Resources/config/form_types.yml
@@ -5,3 +5,5 @@ services:
         public: false
 
     EzSystems\RepositoryForms\Form\Type\SwitcherType: ~
+
+    EzSystems\RepositoryForms\Form\Type\FieldType\BinaryBaseFieldType: ~
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -345,3 +345,6 @@ services:
         class: '%ezrepoforms.form.type.url_edit.class%'
         tags:
             - { name: 'form.type', alias: 'ezrepoforms_url_edit' }
+
+    ezrepoforms.config_resolver.max_upload_size:
+        class: EzSystems\RepositoryForms\ConfigResolver\MaxUploadSize
--- a/lib/FieldType/Mapper/ImageFormMapper.php
+++ b/lib/FieldType/Mapper/ImageFormMapper.php
@@ -14,19 +14,25 @@ use EzSystems\RepositoryForms\FieldType\DataTransformer\ImageValueTransformer;
 use EzSystems\RepositoryForms\FieldType\FieldDefinitionFormMapperInterface;
 use EzSystems\RepositoryForms\FieldType\FieldValueFormMapperInterface;
 use EzSystems\RepositoryForms\Form\Type\FieldType\ImageFieldType;
+use EzSystems\RepositoryForms\ConfigResolver\MaxUploadSize;
 use Symfony\Component\Form\Extension\Core\Type\IntegerType;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\OptionsResolver\Exception\AccessException;
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Validator\Constraints\Range;
 
 class ImageFormMapper implements FieldDefinitionFormMapperInterface, FieldValueFormMapperInterface
 {
     /** @var FieldTypeService */
     private $fieldTypeService;
 
-    public function __construct(FieldTypeService $fieldTypeService)
+    /** @var MaxUploadSize */
+    private $maxUploadSize;
+
+    public function __construct(FieldTypeService $fieldTypeService, MaxUploadSize $maxUploadSize)
     {
         $this->fieldTypeService = $fieldTypeService;
+        $this->maxUploadSize = $maxUploadSize;
     }
 
     public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, FieldDefinitionData $data)
@@ -36,6 +42,16 @@ class ImageFormMapper implements FieldDefinitionFormMapperInterface, FieldValueF
                 'required' => false,
                 'property_path' => 'validatorConfiguration[FileSizeValidator][maxFileSize]',
                 'label' => /** @Desc("Maximum file size (MB)") */ 'field_definition.ezimage.max_file_size',
+                'constraints' => [
+                    new Range([
+                        'min' => 0,
+                        'max' => $this->maxUploadSize->get(MaxUploadSize::MEGABYTES),
+                    ]),
+                ],
+                'attr' => [
+                    'min' => 0,
+                    'max' => $this->maxUploadSize->get(MaxUploadSize::MEGABYTES),
+                ],
             ]);
     }
 
--- a/lib/ConfigResolver/MaxUploadSize.php
+++ b/lib/ConfigResolver/MaxUploadSize.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\ConfigResolver;
+
+class MaxUploadSize
+{
+    const BYTES = 'B';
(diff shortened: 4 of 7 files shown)


ViniTou:
Well, I was in progress on that.
Shouldn't those be handled inside corresponding mappers and fieldTypes together with appropriate constraints?

sunpietro:
@ViniTou the JIRA issue was not in progress. Feel free to update mappers in this PR as I don't have enough knowledge on this topic.

ViniTou:
I will just post it here for review, don't think more time should be spend on that if this is approved

`\EzSystems\RepositoryForms\FieldType\Mapper\MediaFormMapper::mapFieldDefinitionForm`

```
        $fieldDefinitionForm
            ->add('maxSize', IntegerType::class, [
                'required' => false,
                'property_path' => 'validatorConfiguration[FileSizeValidator][maxFileSize]',
                'label' => 'field_definition.ezmedia.max_file_size',
                'constraints' => [
                    new Range([
                        'min' => 0,
                        'max' => $this->getMaxUploadSize()
                    ])
                ],
                'attr' => [
                    'min' => 0,
                    'max' => $this->getMaxUploadSize()
                ]
            ])
```

sunpietro:
ping @webhdx @ViniTou @webhdx

webhdx:
@ViniTou's proposition is actually better. Please remove changes from the twig templates.

ViniTou:
Also, `getMaxUploadSize()` is not available ad hoc in this context, some kind of helper class must be created.

ViniTou:
summon @webhdx 
summon @mikadamczyk as co-defender

;)

webhdx:
summon @Nattfarinn traits :(

ViniTou:
I know, but this just seems like a ideal place to use it.

mikadamczyk:
I am summoned as a defender. In general I do not like traits, but I agree with David. I think this is a good place to use it. Also I am curious why not.

Nattfarinn:
I don't like it. And it's not that I don't like Traits like everyone else because I do (I disagree traits bad pattern just because you can use it in a wrong way, because in such case we wouldn't use PHP in the first place, or in fact any other language at all).

But... I don't like it because all you need is a value injected into various places, and it can be done by Extensions. Once, without need to use `static` and such. Just dynamically set `Container` value in AdminUI extension and inject it wherever you like.

bundle/DependencyInjection/Compiler/MaxUploadSizeParameterPass.php
```diff
@@ -0,0 +1,48 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryFormsBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class MaxUploadSizeParameterPass implements CompilerPassInterface
+{
+    const MAX_UPLOAD_SIZE = 'ezplatform.env.max_upload_size';
```
adamwojs: `ezplatform.env.max_upload_size` will be resolved on container build which in most cases will be resolved based on `php.ini` for CLI environment instead of apache/nginx specific once (upload_max_filesize may differ). 

Another side effect is that changing php.init setting require clearing cache which at least needs to be documented.

ViniTou: Maybe then, we should just drop this max thing and just stick to setting min value (as the task originally says).

bundle/Resources/config/services.yml
```diff
@@ -345,3 +345,6 @@ services:
         class: '%ezrepoforms.form.type.url_edit.class%'
         tags:
             - { name: 'form.type', alias: 'ezrepoforms_url_edit' }
+
+    ezrepoforms.util.max_upload_size:
+        class: EzSystems\RepositoryForms\Util\MaxUploadSize
```
sunpietro: CS: no new line?

lib/FieldType/Mapper/ImageFormMapper.php
```diff
@@ -36,6 +42,16 @@ public function mapFieldDefinitionForm(FormInterface $fieldDefinitionForm, Field
                 'required' => false,
                 'property_path' => 'validatorConfiguration[FileSizeValidator][maxFileSize]',
                 'label' => 'field_definition.ezimage.max_file_size',
+                'constraints' => [
+                    new Range([
+                        'min' => 0,
+                        'max' => $this->maxUploadSize->get(),
```
sunpietro: DB data format inconsistency? That's the reason you're not doing like: `'max' => $this->maxUploadSize->get() / (1024 * 1024),`, right?

ViniTou: ImageFieldType expects input in bytes not megabytes

ViniTou:
@Nattfarinn @webhdx @mikadamczyk @adamwojs 
Can you look at it now, before this goes to QA?

ViniTou:
@barbaragr 
well, `-0` is fine because `0` is. But this derping with `-0.000000000000000000001` is connected to how float is represented as int. It should be done on JS part probably. (to prevent user from entering dot or comma, look how browser dont accept 0.5 but 0.0000000000000000000000001 is fine).

2. It is like that all over the system, it comes from conversion from megabytes to bytes and vice versa, dont think this is in this task scope.

ViniTou:
@Nattfarinn @adamwojs @webhdx 
can we finally move with this? ;)

lib/Util/MaxUploadSize.php
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\Util;
+
+class MaxUploadSize
+{
+    /** @var int */
+    protected $value;
```
adamwojs: CS: properties should be declared after constants 😉

bundle/Resources/config/form_types.yml
```diff
@@ -5,3 +5,4 @@ services:
         public: false
 
     EzSystems\RepositoryForms\Form\Type\SwitcherType: ~
+    EzSystems\RepositoryForms\Form\Type\FieldType\BinaryBaseFieldType: ~
```
webhdx: New line before every new service definition

webhdx:
We were discussing better solution and @Nattfarinn proposed to inject a parameter to the container. New "util" service is an overkill for something this simple.

ViniTou:
@webhdx 
I will just point you to this comment : https://github.com/ezsystems/repository-forms/pull/221#discussion_r174831146

webhdx:
That's ok, @adamwojs sounds resonable.

lib/Util/MaxUploadSize.php
```diff
@@ -0,0 +1,64 @@
...
+            case self::KILOBYTES:
+                $value *= 1 / 1024;
+            case self::BYTES:
+            default:
+        }
+
+        return $value;
+    }
+
+    protected function str2bytes(string $str): int
```
webhdx: Naming is kinda weird on this one. We are using camelCase: `stringToBytes` feels more fitting to our conventions.

webhdx:
Have you considered using more appropriate naming i.e. `MaxFileSizeConfigResolver` or sth instead of `Util` which is something we are trying to avoid?

ViniTou:
@webhdx Fixed and squashed

ViniTou:
@webhdx 
@Nattfarinn 
Can we move with this either way? ;)

barbaragr:
I can't choose, using arrows, value grater than 20 or less than 0 - it's OK. But I can type whatever value I want using keyboard. Is it how it supposed to work? Or maybe if we have a specific range, we shouldn't allow to type for ex. -15?

ViniTou:
@barbaragr 
Is that same as typing non-email in email type input, it can be workarounded on javascript level but that's how `input type = number` works so dunno about it, think this is more frontend or even UI thing.

lserwatka:
@barbaragr we won't bother about it.

barbaragr:
OK, @ViniTou @lserwatka thanks :)

PR #223, opened 2018-03-15
EZP-28944: ISE 500 when deleting Draft

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28944

# Description
Not the solution I'd like to have but I couldn't find any better way to solve the issue. You are expected to return `View` object from the builder where currently we are handling forms. It's not possible to return `RedirectResponse` directly which is the outcome from the `FormProcessors`. I had to hack it a little to encapsulate Response inside `ContentEditSuccessView` and `ContentCreateSuccessView` which is later picked up by `ViewTemplateListener` which is properly handling the Response. Benefit to this resolution is that you can catch the view and render the template if needed.

If anyone has better idea please let me know.

# QA
Test create and edit functionalities.

### DIFF
--- a/bundle/Controller/ContentEditController.php
+++ b/bundle/Controller/ContentEditController.php
@@ -11,7 +11,9 @@ use eZ\Bundle\EzPublishCoreBundle\Controller;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
 use EzSystems\RepositoryForms\Content\View\ContentCreateDraftView;
+use EzSystems\RepositoryForms\Content\View\ContentCreateSuccessView;
 use EzSystems\RepositoryForms\Content\View\ContentCreateView;
+use EzSystems\RepositoryForms\Content\View\ContentEditSuccessView;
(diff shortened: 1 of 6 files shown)


bundle/Controller/ContentEditController.php
```diff
@@ -46,7 +48,19 @@ public function __construct(
      *
      * @return \EzSystems\RepositoryForms\Content\View\ContentCreateView|\Symfony\Component\HttpFoundation\Response
```
ViniTou: is this valid @return doc?

lserwatka:
@Nattfarinn Why hack? Joking or?

bundle/Controller/ContentEditController.php
mikadamczyk: I think yes. Documentation says 'Function can return either of two types'
` /**
  * @return string|null The label's text or null if none provided.
  */
`

ViniTou: Not my point :)

webhdx: Fixed @ViniTou

PR #224, opened 2018-03-16
EZP-28902: Creating a user with existing username results in Internal Server Error.

mikadamczyk:
Jira: https://jira.ez.no/browse/EZP-28902

Depends on: ezsystems/ezpublish-kernel#2279

This PR tries to handle with a new login validation error message from the kernel.

screen shot 2018-03-15 at 11 27 45 am

### DIFF
--- a/lib/Validator/Constraints/FieldValueValidatorMessages.php
+++ b/lib/Validator/Constraints/FieldValueValidatorMessages.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\RepositoryForms\Validator\Constraints;
+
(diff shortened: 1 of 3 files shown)


lib/Validator/Constraints/FieldValueValidatorMessages.php
```diff
@@ -0,0 +1,26 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\RepositoryForms\Validator\Constraints;
+
+use JMS\TranslationBundle\Translation\TranslationContainerInterface;
+use JMS\TranslationBundle\Model\Message;
+
+class FieldValueValidatorMessages implements TranslationContainerInterface
```
alongosz: Please add PhpDoc stating why you need this class so noob like me can understand that as well ;)
I mean, it's used for JMS Translation extractor, right?

mikadamczyk: Done

PR #227, opened 2018-03-22
EZP-28939: Internal Server Error when saving unpublished Draft

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28939

# Description
This PR fixes the issue with unpublished drafts. It was relying on `Location` a lot but unpublished content has no location until published. My PR introduces another property to the `ContentEditView` which tells about `ParentLocation`, it allows `ContentEditView::$location` to be `null`.

### DIFF
--- a/lib/Content/View/Builder/ContentEditViewBuilder.php
+++ b/lib/Content/View/Builder/ContentEditViewBuilder.php
@@ -40,7 +40,7 @@ class ContentEditViewBuilder implements ViewBuilder
     /** @var string */
     private $defaultTemplate;
 
-    /** @var ActionDispatcherInterface */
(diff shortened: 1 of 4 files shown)


lib/Content/View/Builder/ContentEditViewBuilder.php
```diff
@@ -83,16 +83,18 @@ public function buildView(array $parameters)
         $content = $this->resolveContent($parameters, $location, $language);
         $contentType = $this->loadContentType((int) $content->contentInfo->contentTypeId);
         $form = $parameters['form'];
+        $isPublished = null !== $content->contentInfo->mainLocationId && $content->contentInfo->published;
```
mikadamczyk: `$content->contentInfo` is used 4 times here. Can we extract it to variable?

PR #190, opened 2017-12-14
EZP-28527: Fix `ContentTypeFormProcessorTest::testAddFieldDefinition` test

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28527

# Description
Test were broken by #187, this PR fixes failing test. This PR has to be merged up to the `master`.

PR #187, opened 2017-12-12
EZP-28226: User can't add more fields after removing several of them

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28226

# Description
This PR fixes Content Type Edit > Add Field Definition action which had been failing if first field definitions were removed. It was caused by assigning already existing identifier. I created a method which resolves new name making sure it doesn't exist in edited Content Type.

# Remarks
It has to be ported upwards to 1.10 and master.

PR #231, opened 2018-04-09
EZP-28224: Use PHP5.6 compatible function

mnocon:
Done as part of https://jira.ez.no/browse/EZP-28224

This is a follow-up to https://github.com/ezsystems/repository-forms/pull/190, where the tests weren't passing on PHP5.6 (and probably a bug fix to https://github.com/ezsystems/repository-forms/pull/187)

Function `array_column` works on objects only since PHP7 [(Ref),](http://www.php.net/manual/en/function.array-column.php) `$contentTypeDraft->getFieldDefinitions()` returns an array of `FieldDefinition` objects. Changing it to `array_map` should do the trick.

### DIFF
--- a/lib/Form/Processor/ContentTypeFormProcessor.php
+++ b/lib/Form/Processor/ContentTypeFormProcessor.php
@@ -12,9 +12,9 @@ namespace EzSystems\RepositoryForms\Form\Processor;
 
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\Values\ContentType\ContentTypeDraft;
+use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
 use eZ\Publish\API\Repository\Values\ContentType\FieldDefinitionCreateStruct;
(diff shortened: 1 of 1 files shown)


mnocon:
Behat is failing same as in https://github.com/ezsystems/repository-forms/pull/194, I'll try to handle it in next Pull Requests.

lib/Form/Processor/ContentTypeFormProcessor.php
```diff
@@ -170,7 +170,9 @@ private function resolveNewFieldDefinitionIdentifier(
         $startIndex,
         $fieldTypeIdentifier
     ) {
-        $fieldDefinitionIdentifiers = array_column($contentTypeDraft->getFieldDefinitions(), 'identifier');
+        $fieldDefinitionIdentifiers = array_map(function ($fieldDefinition) {
```
alongosz: How about type-hinting callback argument?

lserwatka:
@mnocon you can merge this up if needed

mnocon:
@lserwatka thanks, I'll do it

PR #233, opened 2018-04-10
EZP-28393: Can't set "multi choice" and multiple default in one transaction

mikadamczyk:
Jira: https://jira.ez.no/browse/EZP-28393

### DIFF
--- a/lib/Validator/Constraints/FieldDefinitionDefaultValueValidator.php
+++ b/lib/Validator/Constraints/FieldDefinitionDefaultValueValidator.php
@@ -9,6 +9,7 @@
 namespace EzSystems\RepositoryForms\Validator\Constraints;
 
 use eZ\Publish\API\Repository\Values\ValueObject;
+use eZ\Publish\Core\Repository\Values\ContentType\FieldDefinition;
 use EzSystems\RepositoryForms\Data\FieldDefinitionData;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Validator/Constraints/FieldDefinitionDefaultValueValidator.php
```diff
@@ -51,4 +52,56 @@ protected function generatePropertyPath($errorIndex, $errorTarget)
...
+
+    /**
+     * @param $object
+     *
+     * @return array
+     */
+    private function objectToArray($object): array
+    {
+        $clone = (array)$object;
+        $rtn = array();
```
ViniTou: short array syntax, 
still, don't know how this is nicer than reflection.

ViniTou: Also, not fun of $rtn variable name and I think getUpdatedFieldDefinition() deserves proper doc with explanation.

alongosz: > still, don't know how this is nicer than reflection.

- Performance. Reflections are slow. They can be used in tests (if there's no other way around), never should be used in production code. 
- Reflection is needed when you need to either access private/protected properties or get knowledge on how the object is build. The need for that contradicts DDD and indicates some architecture issues (which indeed we do have)

ViniTou: I know all of that, but tbh I believe that performance won't be an issue here.
And the second point is exactly what this whole thing is about - accessing private property :D

lib/Validator/Constraints/FieldDefinitionDefaultValueValidator.php
```diff
@@ -51,4 +52,39 @@ protected function generatePropertyPath($errorIndex, $errorTarget)
...
+     * @param \eZ\Publish\API\Repository\Values\ValueObject $value
+     *
+     * @return \eZ\Publish\Core\Repository\Values\ContentType\FieldDefinition
+     */
+    private function getUpdatedFieldDefinition(ValueObject $value): FieldDefinition
+    {
+        $oldFieldDef = $value->fieldDefinition;
+
+        $properties = [];
+        foreach ($oldFieldDef->attributes() as $property) {
```
alongosz: This is well defined structure, so just explicitly use properties to initialize new modified value.
`attributes` is deprecated and at some point might disappear.

Note that depending on Core `FieldDefinition` is risky and unsupported, so you should at least add a unit test checking either this specific behavior or `FieldDefinition` structure itself. That way we can at least identify potential issue faster. (I'm more in favor of testing if this Validator returns proper updated structure).

alongosz (changes requested):
Ideally execution of this validator should occur after FieldDefinition gets updated. If you wrap everything with `$repository->beginTransaction()`, `->commit()` and `->rollback()` then it should be fine.

But if that's not possible due to architecture, then things to change:

lib/Validator/Constraints/FieldDefinitionDefaultValueValidator.php
webhdx: Ok I see you copied my solution from slack but I purposely used `getProperties()` instead of `attributes()` because later is deprecated as @alongosz noticed.

alongosz:
@webhdx @mikadamczyk 

> are you aware that the whole transaction happens in the repository? He is not able to (or he should not) alter that from repository forms.

Well... That's weird :)
I've created Gist showing how FieldDefinition with default value should be updated.
https://gist.github.com/alongosz/2e059a9eb1ea9ef2d94b4f15c6843c70

I admit API could use improvement there, like:
- returning updated definition so we can avoid loading it again
- exposing separate API for validating default value

What troubles me mostly is a need to use `Core`, but that's something we have to live with for now (of course Exception can be provided by package as long as it implements API - but still not the best approach).

Please notice in the given example that I've wrapped this code in a single transaction. That way if the value is not valid, rollBack will revert update on settings. This is the way it should be done.

However as I've said before, if that's not possible due to `repository-forms` architecture, then what you should do is to explicitly assign `FieldDefinition` properties and create unit test to be able to track `ezpublish-kernel` possible breaking changes (due to lack of BC promise there).

adamwojs:

PR #236, opened 2018-04-25
EZP-29024: Adding policy user/register to the role throws 500

mikadamczyk:
Jira: https://jira.ez.no/browse/EZP-29024

Now policy `user/register` is a part of kernel. We do not need add this in the repository forms.

### DIFF
--- a/lib/Security/UserRegisterPolicyProvider.php
+++ b/lib/Security/UserRegisterPolicyProvider.php
@@ -11,12 +11,14 @@ use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ConfigBuilde
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Security\PolicyProvider\PolicyProviderInterface;
 
 /**
+ * @deprecated Deprecated since 2.1. No longer used. This policy was moved to the eZ/Publish/Core/settings/policies.yml.
+ *
  * Adds the user/register policy.
  */
 class UserRegisterPolicyProvider implements PolicyProviderInterface
 {
     public function addPolicies(ConfigBuilderInterface $configBuilder)
     {
-        $configBuilder->addConfig(['user' => ['register' => null]]);
+        @trigger_error('Method ' . __METHOD__ . ' is deprecated since 2.1', E_USER_DEPRECATED);
     }
 }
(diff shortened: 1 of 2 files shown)


webhdx:
Removing something is usually risky in terms of BC. If it does not harm I'd probably just deprecate it till 3.0 and then safely remove. @Nattfarinn @alongosz?

lib/Security/UserRegisterPolicyProvider.php
```diff
@@ -1,22 +0,0 @@
...
- */
-namespace EzSystems\RepositoryForms\Security;
-
-use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ConfigBuilderInterface;
-use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Security\PolicyProvider\PolicyProviderInterface;
-
-/**
- * Adds the user/register policy.
- */
-class UserRegisterPolicyProvider implements PolicyProviderInterface
```
adamwojs: As @webhdx said: Let's mark `UserRegisterPolicyProvider` as deprecated and do not register it. Similar changes has been introduced in https://github.com/ezsystems/ezpublish-kernel/pull/2273

webhdx:
@mikadamczyk just to be safe:
```php
trigger_error('Method ' . __METHOD__ . ' is deprecated', E_USER_DEPRECATED);
```

lib/Security/UserRegisterPolicyProvider.php
```diff
@@ -11,12 +11,15 @@
...
+ * @deprecated Deprecated since 2.1. No longer used. This policy was moved to the eZ/Publish/Core/settings/policies.yml.
+ *
  * Adds the user/register policy.
  */
 class UserRegisterPolicyProvider implements PolicyProviderInterface
 {
     public function addPolicies(ConfigBuilderInterface $configBuilder)
     {
+        @trigger_error('Method ' . __METHOD__ . ' is deprecated since 2.1', E_USER_DEPRECATED);
         $configBuilder->addConfig(['user' => ['register' => null]]);
```
alongosz: Maybe let's remove that line only to avoid at the same time users getting hit by missing class and messing up kernel policy matrix. Of course keep deprecation warning.
**But before taking any further action, let's sync on that tomorrow @mikadamczyk @webhdx**

mikadamczyk:
As we discussed I removed method body

PR #240, opened 2018-06-14
EZP-29269: User draft is not discarded on edition cancel

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29269
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    |no (hope so)
| Tests pass?   | yes
| Doc needed?   | no

As disscused, this PR bypass ContentController from flow when editing/creating user.
not sure about race conditions here as draft for user is no longer created when accessing edit page.


bundle/Resources/config/services.yml
```diff
@@ -319,6 +319,9 @@ services:
             - "@ezpublish.api.service.content_type"
             - "@ezpublish.api.service.content"
             - "@ezrepoforms.controller.user"
+            - '@ezpublish.api.inner_repository'
```
alongosz: Please use `@ezpublish.api.repository`

lib/EventListener/UserEditListener.php
```diff
@@ -9,10 +9,15 @@
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\Core\Repository\Repository;
```
alongosz: Using Core Repository will lead to bypassing all the layers. This can't be done in an external package.

lib/EventListener/UserEditListener.php
```diff
@@ -32,31 +39,136 @@ class UserEditListener implements EventSubscriberInterface
...
+     * @throws \Exception
+     */
+    public function beforeUserControllerCreate(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+        if (self::CONTENT_BASE_CREATE_ROUTE === $request->attributes->get('_route')) {
+            $form = $this->formFactory->contentEdit(null, 'content_create');
+            $this->repository->sudo(function () use ($form, $request) {
+                $form->handleRequest($request);
+            });
```
alongosz: :scream: more I read the code, more it scares me. This is not good approach. We use sudo for edge cases only. This PR shouldn't belong to them. Moreover handling entire request is quite a security hole.

lib/EventListener/UserEditListener.php
ViniTou: Undestood, but I strictly need sudo method.

lib/EventListener/UserEditListener.php
ViniTou: Yup, I discussed this with guys in our office, sudo is needed as routing is resolved before authorization.

>Moreover handling entire request is quite a security hole.

how is that different from all other places when we do this? (means, handling whole request).

ViniTou:
As to new routing, wouldn't introducing it be a serious BC break?

lib/EventListener/UserEditListener.php
andrerom: where are we handling whole request in sudo? sudo is mainly meant for specific api calls.

lib/EventListener/UserEditListener.php
andrerom: They are there, you just can't type hint it as it's on purpose left out from interface.

Plan was to split permissions to own layer, but also multi lookup api calls [rfc](https://github.com/ezsystems/ezpublish-kernel/pull/2070) would allow you to read data you don't have access to.

So please type hint API even if it lacks sudo.

lib/EventListener/UserEditListener.php
ViniTou: I can always extract `id` and all other parameters from `$request` but this seems as dirty as this.

ViniTou:
@andrerom 
@alongosz 

Whoa, don't know if it was cache at first try, or some mistake on my part, but I managed to remove Repository dependency.

alongosz:
I kinda forgot about this one. It looks less scary, but still the preferred solution is using different route for editing users to reflect API and avoid any future unforeseen problems.

Moreover we've already used custom action to handle deleting user AFAIR, so this solution was inconsistent with what we had so far.

ViniTou:
@alongosz 
@adamwojs 
@mikadamczyk 
@webhdx 

if you could look at it now, together with https://github.com/ezsystems/ezplatform-admin-ui/pull/588

lib/EventListener/UserEditListener.php
```diff
@@ -32,78 +36,72 @@ class UserEditListener implements EventSubscriberInterface
...
-            $contentId = $routeParams['contentId'];
-            if (null !== $contentId) {
-                try {
-                    $content = $this->contentService->loadContent($contentId);
-                    $contentType = $this->contentTypeService->loadContentType($content->contentInfo->contentTypeId);
-                    $isUserBasedContentType = $this->hasUserField($contentType);
-                } catch (NotFoundException $e) {
-                    return;
-                }
+            if ($form->isValid()) {
```
mikadamczyk: `$form->isValid()` is enough, but we mostly using `$form->isSubmitted() && $form->isValid()` for better readability

ViniTou: There should be both as afair one just isValid is now deprecated :D

webhdx:
Given the amount of hackery we have to introduce proper redirection to user create/edit when `ezuser` based content type is selected, I think it wasn't the best idea. 

It will be less complicated with simple differentiation in AdminUI when user oriented routes should be used. This would make `EzSystems\RepositoryForms\EventListener\UserEditListener` obsolete.

ViniTou:
@webhdx @alongosz 
if you could take a look at it again, both here and in AdminUI.

UserEditListener is still used for all edit buttons in all lists (like in dashboard for example). There is still question if that should be decided in twig template(which url is used), but I am not sure about that.

doc/bc/changes-2.3.md
```diff
@@ -0,0 +1,11 @@
...
+
+Changes affecting version compatibility with former or future versions.
+
+## Removed features
+
+- EZP-29269: Editing content with user content field no longer results with draft creation.
+
+  As part of it :
+   - `\EzSystems\RepositoryForms\EventListener\UserEditListener` was removed 
+   - New routes were introduced `ez_user_create` and `ez_user_update`
```
webhdx: The routes were changed not created.

webhdx:
@ViniTou Just a small nitpick.

lserwatka:
@ViniTou @micszo what is the status here? Feels like this is ready for testing, no?

ViniTou:
@lserwatka 
as soon as https://github.com/ezsystems/ezplatform-admin-ui/pull/588 gets green light, yes.

PR #242, opened 2018-07-20
EZP-29421: Contents of file field are lost if object is saved before publishing

mikadamczyk:
Jira: https://jira.ez.no/browse/EZP-29421

Now when form is sent we set field value only when it was changed.

### DIFF
--- a/tests/RepositoryForms/Form/Processor/ContentFormProcessorTest.php
+++ b/tests/RepositoryForms/Form/Processor/ContentFormProcessorTest.php
@@ -1,0 +1,271 @@
+<?php
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


tests/RepositoryForms/Form/Processor/ContentFormProcessorTest.php
```diff
@@ -0,0 +1,271 @@
...
+        $field2 = $this->getField(self::FIELD_2_IDENTIFIER, self::LANGUAGE_CODE, 'field_2_value');
+
+        $data = new ContentUpdateData([
+            'contentDraft' => $this->contentDraft,
+            'fields' => [],
+            'fieldsData' => [
+                $field1->fieldDefIdentifier => new FieldData([
+                    'field' => $field1,
+                    'fieldDefinition' => $this->getFieldDefinition($field1->fieldDefIdentifier),
+                    'value' => 'new_field_2_value',
```
ViniTou: I think this should have different value. Other than that :+1: if this dosen't mess up with unchecking checkboxes for example.

ViniTou: I think this should have different value.

lserwatka:
@mikadamczyk you can merge this up to 2.2 and master

PR #245, opened 2018-07-26
[WIP] Check tests

mnocon:


mnocon:
The tests here are also broken when running with redis (see second commit). It's the same issue as in https://jira.ez.no/browse/EZP-29291 - redis cache not being cleared when Behat invokes API commands.

PR #247, opened 2018-07-27
[TRAVIS] Switch Travis setup script to fix broken builds

mnocon:
Currently the Travis builds for 2.2 and master branches are failing with:
```
    Given that there is a Content Type with any kind of constraints on a Field Definition # EzSystems\RepositoryForms\Features\Context\ContentEdit::thereIsAContentTypeWithAnyKindOfConstraintsOnAFieldDefinition()
      Database error (RuntimeException)
```
It looks like the install database step is never executed, the following is displayed in log instead:
```
Ok: Connection succeeded to mysql 'ezp'.
[web/ | web/var] is not writable
```

At the same time, the branch 2.2 of ezplatform (which is also running the repository-forms tests) is executing without issues, which made me think it must be something with the install script.

I've decided to switch from using the current (deprecated) script to the newer one, used in AdminUI: https://github.com/ezsystems/ezplatform/blob/master/bin/.travis/trusty/setup_ezplatform.sh

With it the tests are running.

Note: the one failure is a known one and is being currently fixed: https://jira.ez.no/browse/EZP-29091

Also: correcting the value in `composer.json` - repository-forms 2.2 branch should run 2.2 metarepo, not master


mnocon:
@alongosz @m-tyrala I can't add you as reviewers: please have a look.

lserwatka:
@mnocon something is still failing here

mnocon:
@lserwatka the one test failing is the one that uncovered this bug: https://jira.ez.no/browse/EZP-29091 , this failure is expected.

mnocon:
Thanks, I'll merge it upstream

PR #251, opened 2018-08-02
EZP-29290: As an Editor, I want to have preconfigured "Author" section while creating the content

konradoboza:
| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | [EZP-29290](https://jira.ez.no/browse/EZP-29290)
| **Bug/Improvement**| yes
| **Target version** | 2.2
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | yes

JIRA excerpt:
>Currently, in v2 there is no preconfigured author section (fields filled in with the name and e-mail of the currently logged user). It would be more convenient for the editors to have this fields set automatically while creating a new content. Also, prefilling section with the default data should be configurable at the Author FieldType level.

This is a followup for kernel implementation: https://github.com/ezsystems/ezpublish-kernel/pull/2392

**TODO**:
- [x] Implement feature / fix a bug.
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.


konradoboza:
I discovered some problems with saving content with an empty author, some rework here needed.

konradoboza:
I applied changes according to https://github.com/ezsystems/ezpublish-kernel/pull/2417 - default value should be empty to avoid BC break. Changing that also fixed an issue with saving a content with empty Author fields. Please re-review @mikadamczyk @ViniTou @andrerom. :)

PR #265, opened 2018-12-14
Install ezpublish-kernel with @dev

mnocon:
Checking if installing kernel with @dev helps with failing unit tests:
https://travis-ci.org/ezsystems/repository-forms/builds/467887329

AdminUI, which also runs Unit Tests, has kernel set that way already (https://github.com/ezsystems/ezplatform-admin-ui/blob/master/composer.json#L19).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": "^7.1",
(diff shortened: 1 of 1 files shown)


mnocon:
@lserwatka this can be merged

PR #266, opened 2018-12-18
EZEE-2132: User should be redirected to created Landing Page instead of it's parent

Nattfarinn:
JIRA: https://jira.ez.no/browse/EZEE-2132

### DIFF
--- a/lib/Form/Processor/ContentFormProcessor.php
+++ b/lib/Form/Processor/ContentFormProcessor.php
@@ -89,6 +89,9 @@ class ContentFormProcessor implements EventSubscriberInterface
         $referrerLocation = $event->getOption('referrerLocation');
         $contentLocation = $this->resolveLocation($draft, $referrerLocation, $data);
 
+        $event->setPayload('content', $draft);
+        $event->setPayload('is_new', $draft->contentInfo->isDraft());
+
         $defaultUrl = $this->router->generate('ez_content_draft_edit', [
             'contentId' => $draft->id,
             'versionNo' => $draft->getVersionInfo()->versionNo,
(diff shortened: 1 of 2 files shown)


lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -117,6 +117,10 @@ public function processPublish(FormActionEvent $event)
         $draft = $this->saveDraft($data, $form->getConfig()->getOption('languageCode'));
         $content = $this->contentService->publishVersion($draft->versionInfo);
 
+        $event->setPayload('contentDraft', $draft);
```
webhdx: I'd name it `content` and do not pass `versionInfo` and `contentInfo` because they can be obtained from `content` directly.

lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -151,11 +155,17 @@ public function processCancel(FormActionEvent $event)
         $contentInfo = $content->contentInfo;
         $versionInfo = $data->contentDraft->getVersionInfo();
 
+        $event->setPayload('contentDraft', $content);
```
webhdx: Same as above

lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -186,6 +196,10 @@ public function processCreateDraft(FormActionEvent $event)
         $contentDraft = $this->contentService->createContentDraft($contentInfo, $versionInfo);
         $referrerLocation = $event->getOption('referrerLocation');
 
+        $event->setPayload('contentDraft', $contentDraft);
```
webhdx: How does `contentDraft` relate to `versionInfo`? Is it the same version? If so I'd limit payloads to `content` only.

lserwatka:
@Nattfarinn you can merge it up

PR #267, opened 2018-12-20
EZP-29925: Content Type Translate in Edit mode is not based on browser's settings

ViniTou:
> JIRA: https://jira.ez.no/browse/EZP-29925


lib/FieldType/Mapper/AuthorFormMapper.php
```diff
@@ -60,16 +63,14 @@ public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
         $fieldSettings = $fieldDefinition->getFieldSettings();
         $formConfig = $fieldForm->getConfig();
         $names = $fieldDefinition->getNames();
-        $label = $fieldDefinition->getName($formConfig->getOption('languageCode'))
```
alongosz: Uh :)
Actually, assuming you've obtained that `FieldDefinition` from a Content Type loaded via `loadContentType($contentTypeId, $languageCodes);`, where `$languageCode` comes from `UserLanguagePreferenceProvider`, the call to `$fieldDefinition->getName()` should get you a proper label. 

This code should not have been here in the first place, because it repeats internal Repository's logic.

lib/FieldType/TranslatableLabel.php
```diff
@@ -0,0 +1,23 @@
...
+
+trait TranslatableLabel
+{
+    private function resolveLabel(array $names, array $languageCodes): string
+    {
+        foreach ($languageCodes as $languageCode) {
+            if (array_key_exists($languageCode, $names)) {
+                return $names[$languageCode];
+            }
+        }
```
alongosz: This definitely cannot be implemented like that, because it also duplicates Repository's internal logic. ContentType needs to be loaded with proper language code and everything should be fine.

So please check if that already happens (since I see in the diffs here that you're actually injecting proper language code) but if not, please trace back where the loading of `FieldDefinition`s occur and inject proper language there.

Keep in mind that this needs to work with 3rd party Field Types.

lib/Content/View/Filter/ContentCreateViewFilter.php
```diff
@@ -31,20 +32,25 @@ class ContentCreateViewFilter implements EventSubscriberInterface
 
     /** @var \Symfony\Component\Form\FormFactoryInterface */
     private $formFactory;
+    /** @var \eZ\Publish\Core\MVC\Symfony\Locale\UserLanguagePreferenceProvider */
```
adamwojs: Nitpick: Missing new line

lib/Content/View/Builder/ContentCreateViewBuilder.php
```diff
@@ -187,21 +193,21 @@ private function resolveLanguage(array $parameters): Language
 
     /**
      * @param array $parameters
-     * @param \eZ\Publish\API\Repository\Values\Content\Language $language
+     * @param array $languageCodes
```
adamwojs:
```suggestion
      * @param string[] $languageCodes
```

lib/Content/View/Builder/ContentCreateViewBuilder.php
```diff
@@ -187,21 +193,21 @@ private function resolveLanguage(array $parameters): Language
...
      *
      * @return \eZ\Publish\API\Repository\Values\ContentType\ContentType
      *
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
-     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentException
```
adamwojs: `\eZ\Publish\API\Repository\Exceptions\InvalidArgumentException` was fine.

lib/Content/View/Filter/ContentCreateViewFilter.php
```diff
@@ -105,17 +115,19 @@ private function resolveContentCreateData(
      * @param \EzSystems\RepositoryForms\Data\Content\ContentCreateData $contentCreateData
      * @param string $languageCode
      *
-     * @return \Symfony\Component\Form\FormInterface
+     * @param array $preferredLanguages
```
adamwojs: Probably should be

```suggestion
     * @param string[] $preferredLanguages
```

 if i'm wrong, feel free to ignore this comment

PR #270, opened 2019-01-02
EZP-29979: As a developer I would like to use getLocation method on ContentCreateSuccessView and ContentEditSuccessView

kmadejski:
JIRA: https://jira.ez.no/browse/EZP-29979

### DIFF
--- a/lib/Content/View/ContentCreateSuccessView.php
+++ b/lib/Content/View/ContentCreateSuccessView.php
@@ -7,12 +7,17 @@ declare(strict_types=1);
 
 namespace EzSystems\RepositoryForms\Content\View;
 
+use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\Core\MVC\Symfony\View\BaseView;
+use eZ\Publish\Core\MVC\Symfony\View\LocationValueView;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Controller\ControllerReference;
 
-class ContentCreateSuccessView extends BaseView
+class ContentCreateSuccessView extends BaseView implements LocationValueView
 {
+    /** @var \eZ\Publish\API\Repository\Values\Content\Location|null */
+    private $location;
+
     /**
      * @param \Symfony\Component\HttpFoundation\Response $response
      *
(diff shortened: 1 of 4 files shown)


lib/Content/View/ContentCreateSuccessView.php
```diff
@@ -25,4 +30,20 @@ public function __construct(Response $response)
         $this->setResponse($response);
         $this->setControllerReference(new ControllerReference('ez_content_edit:createWithoutDraftSuccessAction'));
     }
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\Location|null $location
+     */
+    public function setLocation(?Location $location)
```
ViniTou: Not sure if there is any case when this is set to null

kmadejski: Yep, just wanted to be consistent with https://github.com/ezsystems/repository-forms/blob/master/lib/Content/View/ContentEditView.php#L61, but it seems that indeed there is no case for that. Let's wait for @webhdx 🙂

andrerom: Only hypothetical case would be when you want to remove location.

For getLocation it's needed for cases where there are no locations _(most notably draft, but also technically possible with location less content)_, but I also don't see a strong need to support this on setLocation().

lib/Content/View/ContentCreateSuccessView.php
```diff
@@ -25,4 +30,20 @@ public function __construct(Response $response)
         $this->setResponse($response);
         $this->setControllerReference(new ControllerReference('ez_content_edit:createWithoutDraftSuccessAction'));
     }
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location
+     */
+    public function setLocation(Location $location)
```
webhdx:
```suggestion
    public function setLocation(Location $location): void
```

lib/Content/View/ContentEditSuccessView.php
```diff
@@ -25,4 +30,20 @@ public function __construct(Response $response)
         $this->setResponse($response);
         $this->setControllerReference(new ControllerReference('ez_content_edit:editVersionDraftSuccessAction'));
     }
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location
+     */
+    public function setLocation(Location $location)
```
webhdx:
```suggestion
    public function setLocation(Location $location): void
```

kmadejski:
@andrerom / @ViniTou I have to allow passing `null` as `$location` due to the issue found by QA. 

@mnocon could you please retest?

kmadejski:
@mnocon do you see any chance to give it another round of testing? 🙂

mnocon:
@kmadejski there is a chance 😉 But you're probably aware that 60% of our QA team is on sick leave and it complicates things a bit

kmadejski:
Good question @mnocon, honestly, I don't know. @ViniTou / @webhdx / @Nattfarinn / @andrerom 
 WDYT?

PR #271, opened 2019-01-14
EZP-30008: More unique uniqid()

glye:
Implements https://jira.ez.no/browse/EZP-30008

Use `prefix` and `more_entropy` wherever possible.

See also
https://github.com/ezsystems/ezpublish-kernel/pull/2523
https://github.com/ezsystems/BehatBundle/pull/72


glye:
Fixup: Testing to see why failures are happening.

glye:
Fixup: WTF, it passed! Reverting to see if there was a glitch in the matrix.

PR #273, opened 2019-01-14
EZP-29849: Allow user to remove location in subtree limitation

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-29849

## Description

Changes required to make Location object available during rendering edit form of the location based limitations.  

In the limitation template (e.g. `src/bundle/Resources/views/Limitation/udw_limitation_value.html.twig` in `ezplatform-admin-ui`) location instances are available using `form.limitationValues.vars.data`. The model/view format of data is not changed:

zrzut ekranu z 2019-01-14 20-32-00


lib/Limitation/DataTransformer/UDWBasedValueModelTransformer.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function reverseTransform($value)
+    {
+        if (!is_array($value)) {
+            return null;
+        }
+
+        return array_map(function (Location $location) {
```
ViniTou: don't know if this even matters, but that could be replaced with array_column.

adamwojs:
PR updated according to @ViniTou suggestions.

PR #275, opened 2019-01-18
Remove dot from Content Type identifier

mnocon:
Similar as in https://github.com/ezsystems/BehatBundle/pull/73 - the dot cannot be used in Content Type identifier.


glye:
@mnocon I thought Travis would catch this, apparently not. FWIW, the kernel commits have passed both Travis and AppVeyor.

PR #280, opened 2019-02-27
EZP-30181: Editor is not able to select any item in the ezobjectrelation/ezobjectrelationlist FTs if starting location is not accessible

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-30181

## Description 

Follow up for https://github.com/ezsystems/repository-forms/pull/279. Editor is not able to select any content item as value of the `ezobjectrelation`/`ezobjectrelationlist` field if the starting location is not accessible for some reason (location was trashed or because of lack of permission). In such case we will fallback starting location to null witch is technically content tree root. 

In such case we assume that content tree root should be used as starting point for selection.


lserwatka:
Could you merge it up?

adamwojs:
Done.

PR #293, opened 2019-05-21
EZP-30581: Upgrade `ezsystems/repository-forms` to use Symfony 4 components

webhdx:


.travis.yml
```diff
@@ -6,25 +6,21 @@ env:
...
+    - name: "[PHP 7.3] Repository-forms tests on Clean P"
+      php: 7.3
       env:
         - BEHAT_OPTS="--mode=behat --profile=repository-forms --non-strict"
-        - PHP_IMAGE=ezsystems/php:7.1-v1
-    # 7.2
-    - name: "[PHP 7.2] PHP Unit tests"
-      php: 7.2
-      env: TEST_CONFIG="phpunit.xml.dist"
+        - PHP_IMAGE=ezsystems/php:7.3
```
ViniTou: isnt correct image name `ezsystems/php:7.3-v1`? 
@mnocon

mnocon: Yep, usually we specify it that way (example: https://github.com/ezsystems/ezplatform/blob/master/.env#L13 )

webhdx: Thanks, fixed

PR #294, opened 2019-05-23
EZP-30586: Adding of Selection FieldType without configured options leads to an 500 error

kmadejski:
JIRA: https://jira.ez.no/browse/EZP-30586


andrerom:
@lserwatka Definitely on the 2.5.2 wish list :)

kmadejski:
@lserwatka done in https://github.com/ezsystems/repository-forms/commit/42f8d8e2039e2776ccb958a6e43f18b6744f0a7c.

PR #295, opened 2019-05-24
EZP-30539: Move Platform Behat Bundle to BehatBundle #83

mnocon:
Missing PR for https://jira.ez.no/browse/EZP-30539 , adjusting tests for the lack of PlatformBehatBundle.

This is not the cleanest of fixes, as I've missed this during https://jira.ez.no/browse/EZP-30539 - I don't think we should spend more time on this, given that repository-forms will be removed in 3.0 and the Contexts here will have to be ported and aligned to current approach in BehatBundle (some parts of that already happened).


mnocon:
@lserwatka this one can be merged

PR #301, opened 2019-07-01
EZP-30712: Fixed reverseTransform for midnight value

SerheyDolgushev:
**JIRA**: https://jira.ez.no/browse/EZP-30712

Can't set a time to be midnight without this fix

### DIFF
--- a/lib/FieldType/DataTransformer/TimeValueTransformer.php
+++ b/lib/FieldType/DataTransformer/TimeValueTransformer.php
@@ -50,7 +50,7 @@ class TimeValueTransformer implements DataTransformerInterface
      */
     public function reverseTransform($value)
     {
-        if (is_null($value) || $value === '') {
+        if (null === $value || $value === '') {
             return null;
         }
 
--- a/lib/FieldType/DataTransformer/TimeValueTransformer.php
+++ b/lib/FieldType/DataTransformer/TimeValueTransformer.php
@@ -50,7 +50,7 @@ class TimeValueTransformer implements DataTransformerInterface
      */
     public function reverseTransform($value)
     {
-        if (empty($value)) {
+        if (is_null($value) || $value === '') {
             return null;
         }
 
--- a/tests/RepositoryForms/FieldType/DataTransformer/TimeValueTransformerTest.php
+++ b/tests/RepositoryForms/FieldType/DataTransformer/TimeValueTransformerTest.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * This file is part of the eZ RepositoryForms package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\RepositoryForms\Tests\FieldType\DataTransformer;
+
+use DateTime;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+use eZ\Publish\Core\FieldType\Time\Value;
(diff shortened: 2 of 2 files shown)


micszo:
Hi @SerheyDolgushev, thanks for opening the PR! Could you change base and rebase to 2.5?

lib/FieldType/DataTransformer/TimeValueTransformer.php
```diff
@@ -50,7 +50,7 @@ public function transform($value)
      */
     public function reverseTransform($value)
     {
-        if (empty($value)) {
+        if (is_null($value) || $value === '') {
```
alongosz: CS:
```suggestion
        if (null === $value || $value === '') {
```

SerheyDolgushev: Done

SerheyDolgushev:
Requested CS changes and rebase against 2.5 is done.

`TimeValueTransformer` uses `eZ\Publish\Core\FieldType\Time\Value`, which is not the part of this package. So there is no easy way to test `TimeValueTransformer`. @alongosz can you please suggest a solution here?

alongosz:
This is still unresolved:

> @SerheyDolgushev is it possible to cover this with some minimal unit tests? I'd like to know what empty-ish value occurs for midnight w/o debugging it myself (good input data for test case) ;)
> 
> Seems we're missing tests for `TimeValueTransformer` so please add minimal coverage of just `testTransform` with a few inputs. Please see how it's done in `\EzSystems\RepositoryForms\Tests\FieldType\DataTransformer\DateIntervalToArrayTransformerTest`.

I'm sorry, but I don't have time to reproduce this myself and decide if the fix is correct based on behavior. This is why tests are helpful.

SerheyDolgushev:
@alongosz please review https://github.com/ezsystems/repository-forms/pull/301/commits/0e9d3b3f7a837272646596bb8745df8c3c0148f2.

tests/RepositoryForms/FieldType/DataTransformer/TimeValueTransformerTest.php
```diff
@@ -0,0 +1,52 @@
...
+        $transformer = new TimeValueTransformer();
+
+        self::assertNull($transformer->transform($value));
+    }
+
+    public function testTransformInvalidValue()
+    {
+        $transformer = new TimeValueTransformer();
+
+        self::expectException(TransformationFailedException::class);
```
alongosz: AFAICS `expectException` is not static, so should be refered by `$this->`

tests/RepositoryForms/FieldType/DataTransformer/TimeValueTransformerTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+        self::assertSame($time, $transformer->transform($value));
+    }
+
+    public function testTransformZero()
+    {
+        $value = new Value(0);
+        $transformer = new TimeValueTransformer();
+
+        self::assertSame(0, $transformer->transform($value));
```
alongosz: Ok, now I get the fix, thanks!

tests/RepositoryForms/FieldType/DataTransformer/TimeValueTransformerTest.php
SerheyDolgushev: Done: https://github.com/ezsystems/repository-forms/pull/301/commits/da92b939fe773a623e39b5d9f3775082aedebab8

lserwatka:
@alongosz could you merge it up?

alongosz:
@lserwatka done.

Thank you @SerheyDolgushev :tada:

PR #302, opened 2019-07-16
EZP-30529: Injected Field Types into Repository from DI Container

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-30529

Aligns to changes done in https://github.com/ezsystems/ezpublish-kernel/pull/2688

### DIFF
--- a/lib/Form/Type/ContentType/FieldTypeChoiceType.php
+++ b/lib/Form/Type/ContentType/FieldTypeChoiceType.php
@@ -7,7 +7,7 @@
  */
 namespace EzSystems\RepositoryForms\Form\Type\ContentType;
 
-use eZ\Publish\Core\Base\Container\ApiLoader\FieldTypeCollectionFactory;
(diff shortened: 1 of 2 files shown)


lib/Form/Type/ContentType/FieldTypeChoiceType.php
```diff
@@ -64,7 +64,7 @@ public function getParent()
     private function getFieldTypeChoices()
     {
         $choices = [];
-        foreach ($this->fieldTypeCollectionFactory->getConcreteFieldTypesIdentifiers() as $fieldTypeIdentifier) {
+        foreach ($this->fieldTypeRegistry->getConcreteFieldTypesIdentifiers() as $fieldTypeIdentifier) {
```
alongosz: Repo-forms is kinda external package, so I'd rather use `\eZ\Publish\API\Repository\FieldTypeService` here.

Though it can be considered as out of scope, because we used FT Collection Factory before, which had the same issue.

PR #305, opened 2019-08-04
EZP-30813: Stop using deprecated '_ezpublishLocation' route

wizhippo:
Repository forms uses '_ezpublishLocation' route which is deprecated.

If using repository forms on front end it causes a redirect to a page which the front end cannot access.

Instead using '_ezpublishLocation' we should use it's replacement or 'ez_urlalias' so the redirect after publish goes the to expected location for the siteaccess

### DIFF
--- a/lib/Form/Processor/ContentFormProcessor.php
+++ b/lib/Form/Processor/ContentFormProcessor.php
@@ -125,7 +125,7 @@ class ContentFormProcessor implements EventSubscriberInterface
         $event->setPayload('is_new', $draft->contentInfo->isDraft());
 
         $redirectUrl = $form['redirectUrlAfterPublish']->getData() ?: $this->router->generate(
-            '_ezpublishLocation', [
+            'ez_urlalias', [
                 'locationId' => $content->contentInfo->mainLocationId,
             ]
         );
@@ -146,7 +146,7 @@ class ContentFormProcessor implements EventSubscriberInterface
 
         if ($data->isNew()) {
             $response = new RedirectResponse($this->router->generate(
-                '_ezpublishLocation',
+                'ez_urlalias',
                 ['locationId' => $data->getLocationStructs()[0]->parentLocationId]
             ));
             $event->setResponse($response);
@@ -171,7 +171,7 @@ class ContentFormProcessor implements EventSubscriberInterface
         }
 
         $url = $this->router->generate(
-            '_ezpublishLocation',
+            'ez_urlalias',
             ['locationId' => $redirectionLocationId],
             UrlGeneratorInterface::ABSOLUTE_URL
         );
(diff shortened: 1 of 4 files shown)


webhdx:
This change is good as long as AdminUI still uses `_ezpublishLocation` route. Did you check it?

wizhippo:
@webhdx No admin uses the updated 'ez_urlalias' but is working.  Why would admin need to use '_ezpublishLocation'?  Should aliases not work in admin too?

adamwojs:
FYI: @wizhippo Problem reported by failed build (https://travis-ci.org/ezsystems/repository-forms/jobs/567439386) has been fixed by https://github.com/ezsystems/repository-forms/pull/308 so rebase will be needed here.

webhdx:
@wizhippo In AdminUI we use `_ezpublishLocation` extensively. We intentionally did not want URL aliases to be used by administration panel as you most likely expect "technical URLs" to i.e. quickly change ID to get to the desired content. With your change content editing would use URL aliases and all other areas will still rely on `_ezpublishLocation` (subitems, UDW etc.).

wizhippo:
@webhdx Ok I understand, but then what should it be replaced with since it is deprecated? Also we then should provide an option (or at lease a solution) so it can also be used on the front end too.  To me this package looks like it should a usable in both places.  If not that should be documented too.  To clarify if you try to use this as is without this change, after content is created it redirect to a location like '/content/location/1836' as you know.  These are not allowed on the fronted and cause an exception.  It would suck to have to add our own form processors to override the ones in this bundle.

pspanja:
@webhdx 

> In AdminUI we use _ezpublishLocation extensively. We intentionally did not want URL aliases to be used by administration panel as you most likely expect "technical URLs" to i.e. quickly change ID to get to the desired content. With your change content editing would use URL aliases and all other areas will still rely on _ezpublishLocation (subitems, UDW etc.).

You should allow both, with URL alias being the default. That would be the same as in Legacy admin UI, with `content/view/full/<location id>` route. It's not a good experience when you are not able to find the frontend URL for a Content when viewing it in the admin UI.

emodric:
> It's not a good experience when you are not able to find the frontend URL for a Content when viewing it in the admin UI.

Vice versa also applies. You should be able to quickly locate the content in admin based on frontend URL, just by replacing the siteaccess in the URL.

pspanja:
> Vice versa also applies. You should be able to quickly locate the content in admin based on frontend URL, just by replacing the siteaccess in the URL.

That actually already works, the problem is `_ezpublishLocation` being the default in admin.

emodric:
Good to know :+1:

wizhippo:
Just to clarify that in backend both alias and location work as discussed above.  The issue really is that after content is published the redirect goes to `_ezpublishLocation` instead of alias which is fine for backend but for frontend is bad.

wizhippo:
Perhaps something like the below, though the form processors would then have to be site aware.

        $route = 'ez_urlalias';
        if ($isAdminSiteAccess) {
            $route = '_ezpublishLocation';
        }

There are even more issues such as 'Preview' is also added to forms, admin-ui adds this via a type extension regardless if frontend/backend.  Just more example of how it it more apparent this repo is really meant for backend. This is fine but this should really be admin-repository-forms instead so it is clear it's not really meant for use elsewhere.

PR #315, opened 2019-09-11
EZP-30938: Removed usage of deprecated setCurrentUser method

ViniTou:
> Jira: https://jira.ez.no/browse/EZP-30938

### DIFF
--- a/behat_suites.yml
+++ b/behat_suites.yml
@@ -7,9 +7,7 @@ repository-forms:
                 - vendor/ezsystems/repository-forms/features/ContentEdit
             contexts:
                 - EzSystems\RepositoryForms\Behat\Context\ContentEditContext
-                - EzSystems\RepositoryForms\Behat\Context\ContentTypeContext:
-                    permissionResolver: '@eZ\Publish\API\Repository\PermissionResolver'
-                    contentTypeService: '@ezpublish.api.service.content_type'
+                - EzSystems\RepositoryForms\Behat\Context\ContentTypeContext
                 - EzSystems\RepositoryForms\Behat\Context\PagelayoutContext
                 - EzSystems\Behat\Browser\Context\Hooks
         fieldtype_form:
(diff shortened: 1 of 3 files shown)


lib/Behat/Context/ContentTypeContext.php
```diff
@@ -44,7 +44,7 @@ final class ContentTypeContext extends RawMinkContext implements Context, Snippe
     public function __construct(Repository $repository, ContentTypeService $contentTypeService)
     {
         $this->repository = $repository;
-        $this->repository->setCurrentUser(new UserReference($this->adminUserId));
+        $this->repository->getPermissionResolver()->setCurrentUserReference(new UserReference($this->adminUserId));
```
alongosz: Is it possible to inject PermissionResolver directly?

ViniTou:
@alongosz 
done, but I have this feeling that this must be repeated for every suits that is using this conext, I am right @mnocon

behat_suites.yml
```diff
@@ -7,21 +7,29 @@ repository-forms:
                 - vendor/ezsystems/repository-forms/features/ContentEdit
             contexts:
                 - EzSystems\RepositoryForms\Behat\Context\ContentEditContext
-                - EzSystems\RepositoryForms\Behat\Context\ContentTypeContext
+                - EzSystems\RepositoryForms\Behat\Context\ContentTypeContext:
+                    permissionResolver: '@eZ\Publish\API\Repository\PermissionResolver'
```
alongosz: This wasn't here before, even though Context already had arguments, so I didn't expect it's needed.

ViniTou: I was rather thinking about things like https://github.com/ezsystems/ezplatform-admin-ui/blob/master/behat_suites.yml#L36 as when removed dependency by annotations this is no longer built properly.

ViniTou: as for one line change, injection with annotations does not work correctly (or I just couldn't find a way with service names using FCQN).

mnocon: @ViniTou it's fixed by https://github.com/ezsystems/ezpublish-kernel/pull/2750

If you can wait a bit so I can merge it then using that would be the preffered approach, as keeping Context definitions in multiple behat_suites file is a tiresome thing.

ViniTou:
This will be solved using @injectService when https://github.com/ezsystems/ezpublish-kernel/pull/2750 is merged :)

behat_suites.yml
```diff
@@ -14,7 +14,7 @@ repository-forms:
             paths:
                 - vendor/ezsystems/repository-forms/features/FieldTypeForm
             contexts:
-                - EzSystems\RepositoryForms\Behat\Context\ContentTypeContext
+                - EzSystems\RepositoryForms\Behat\Context\ContentTypeContext:
```
mnocon: Unneeded change 😉

ViniTou:
```suggestion
                - EzSystems\RepositoryForms\Behat\Context\ContentTypeContext
```

ViniTou: fixd

alongosz:
Behat-related changes only (and passing), no further QA required,

PR #316, opened 2019-09-24
EZP-30313 As an editor I want to add more than 3 digits after decimal…

aleblanc:
… point in a float field type

The NumberType symfony scale value is 3 by default. We can allow 20 decimal by default to avoid losing decimals.

### DIFF
--- a/lib/Form/Type/FieldType/FloatFieldType.php
+++ b/lib/Form/Type/FieldType/FloatFieldType.php
@@ -65,7 +65,7 @@ class FloatFieldType extends AbstractType
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver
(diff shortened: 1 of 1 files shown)


webhdx:
Hello @aleblanc 
This looks like very specific to your use case. By default Symfony uses scale from configured locale. This is probably what most users want. Your idea should be implemented on different layer though as Repository Forms package works on top of FieldType implementation in ezsystems/ezpublish-kernel. It should be configurable there and Repository Forms would respect that to display proper input form. 

Alternatively you can make Type Extension (https://symfony.com/doc/current/form/create_form_type_extension.html) and fix the issue in your project.

adamwojs:

PR #319, opened 2019-10-02
EZP-30983: As an Administrator, I want prevent users from reusing old password

adamwojs:
> JIRA:  https://jira.ez.no/browse/EZP-30983
> Requires: https://github.com/ezsystems/ezpublish-kernel/pull/2788

## Description

See https://github.com/ezsystems/ezpublish-kernel/pull/2788


lserwatka:
You can merge it up.

PR #324, opened 2019-10-30
EZP-30997: fixed adding translation by an user with language limitation set

pawbuj:
jira : https://jira.ez.no/browse/EZP-30997

### DIFF
--- a/lib/Form/Processor/ContentFormProcessor.php
+++ b/lib/Form/Processor/ContentFormProcessor.php
@@ -119,7 +119,7 @@ class ContentFormProcessor implements EventSubscriberInterface
 
         $draft = $this->saveDraft($data, $form->getConfig()->getOption('languageCode'));
         $versionInfo = $draft->versionInfo;
-        $content = $this->contentService->publishVersion($versionInfo, [$versionInfo->initialLanguageCode]);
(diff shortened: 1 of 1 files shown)


lib/Form/Processor/ContentFormProcessor.php
```diff
@@ -119,7 +119,7 @@ public function processPublish(FormActionEvent $event)
 
         $draft = $this->saveDraft($data, $form->getConfig()->getOption('languageCode'));
         $versionInfo = $draft->versionInfo;
-        $content = $this->contentService->publishVersion($versionInfo, [$versionInfo->initialLanguageCode]);
+        $content = $this->contentService->publishVersion($versionInfo);
```
alongosz: This will most likely break [EZP-29694](https://jira.ez.no/browse/EZP-29694) fixed via #296. As stated in ezsystems/ezpublish-kernel#2846 the issue is probably caused by missing target there.

alongosz (changes requested):

lib/Form/Processor/ContentFormProcessor.php
pawbuj: issue fixed in the kernel https://github.com/ezsystems/ezpublish-kernel/pull/2858

PR #325, opened 2019-10-31
EZP-28900: Removed view transformer from FloatFieldType

pawbuj:
jira: https://jira.ez.no/browse/EZP-28900

The root of the issue is that our `FloatFieldType` uses view transformer : `NumberToLocalizedStringTransformer`. This transformer is not set explicitly but it is inherited from symfony's `NumberType`. Symfony NumberType needs this transformer because  default `form_fields.html.twig` use `<input type='text">`  to display `number_widget`. AdminUI overwrites symfony's `form_fields.html.twig` including `number_widget` and number are displayed as `<input type="number"`. So this transformer is no needed and causes problems described in the ticket.  

related to : https://github.com/ezsystems/ezpublish-kernel/pull/2849

### DIFF
--- a/lib/Form/Type/FieldType/FloatFieldType.php
+++ b/lib/Form/Type/FieldType/FloatFieldType.php
@@ -45,6 +45,8 @@ class FloatFieldType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder->addModelTransformer(new FieldValueTransformer($this->fieldTypeService->getFieldType('ezfloat')));
+
+        // Removes NumberToLocalizedStringTransformer which breaks "number" type HTML input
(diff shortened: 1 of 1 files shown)


katarzynazawada:
Now, float is shown in edit mode but there is a difference between displaying decimal separators. In preview is `.` and in editing `,` . Not sure if it is related to reported customer request or it should be reported separately as follow-up.  //cc @kmadejski

webhdx:
@katarzynazawada I've fixed it under the same ticket but it required changes in AdminUI: https://github.com/ezsystems/ezplatform-admin-ui/pull/1165

kmadejski:
👍 for fix proposed by @webhdx

lserwatka:
Could you merge it up?

PR #326, opened 2019-11-12
EZP-30882: As a developer I want to use custom Field Type as Content thumbnail

Nattfarinn:
# Depends on: https://github.com/ezsystems/ezpublish-kernel/pull/2868

### DIFF
--- a/bundle/Resources/config/services.yaml
+++ b/bundle/Resources/config/services.yaml
@@ -93,7 +93,10 @@ services:
 
     ezrepoforms.field_definition.form_type:
         class: "%ezrepoforms.field_definition.form_type.class%"
-        arguments: ["@ezrepoforms.field_type_form_mapper.dispatcher", "@ezpublish.api.service.field_type"]
+        arguments:
+            - "@ezrepoforms.field_type_form_mapper.dispatcher"
+            - "@ezpublish.api.service.field_type"
+            - '@eZ\Publish\Core\Repository\Strategy\ContentThumbnail\Field\ContentFieldStrategy'
         calls:
             - [setGroupsList, ["@ezpublish.fields_groups.list"]]
         tags:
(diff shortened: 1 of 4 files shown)


lib/Form/Type/FieldDefinition/FieldDefinitionType.php
```diff
@@ -41,11 +44,16 @@ class FieldDefinitionType extends AbstractType
      * @var FieldsGroupsList
      */
     private $groupsList;
+    /**
```
ViniTou: new line

ViniTou (approved):

bdunogier:
I'd change the commit message to what the patch actually does, but I'm not a developer anymore ;)

`Integrated the "isThumbnail" field definition option (EZP-30882)`

bundle/Resources/config/services.yaml
```diff
@@ -93,7 +93,10 @@ services:
 
     ezrepoforms.field_definition.form_type:
         class: "%ezrepoforms.field_definition.form_type.class%"
-        arguments: ["@ezrepoforms.field_type_form_mapper.dispatcher", "@ezpublish.api.service.field_type"]
+        arguments:
+            - "@ezrepoforms.field_type_form_mapper.dispatcher"
+            - "@ezpublish.api.service.field_type"
+            - '@eZ\Publish\Core\Repository\Strategy\ContentThumbnail\Field\ContentFieldStrategy'
```
alongosz: Please create a service in Kernel which uses interface (API/SPI?) and use it here.

alongosz (changes requested):
> I'd change the commit message to what the patch actually does, but I'm not a developer anymore ;)
> 
> `Integrated the "isThumbnail" field definition option (EZP-30882)`

Definitely, I need to distinguish this on the list of changes and later on when looking for something.
Though it should rather be:
> EZP-30882: Integrated the "isThumbnail" field definition option

Other remark:

Nattfarinn:
Repository Forms is deprecated

PR #331, opened 2020-01-09
EZP-31277: Added server timezone offset attr to ezdate form field

barw4:
Jira: https://jira.ez.no/browse/EZP-31277

flatpickr widget is instantiating with the wrong timezone offset, it should be server's timezone, not browser's one. Such configuration may result in an invalid default date for _ezdate_ field.

Added additional timezone offset attribute to _ezdate_ field in order to get this data later by flatpickr widget.

Related PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/1185

### DIFF
--- a/lib/FieldType/Mapper/DateFormMapper.php
+++ b/lib/FieldType/Mapper/DateFormMapper.php
@@ -60,7 +60,7 @@ class DateFormMapper implements FieldDefinitionFormMapperInterface, FieldValueFo
                         [
                             'required' => $fieldDefinition->isRequired,
                             'label' => $fieldDefinition->getName(),
-                            'attr' => ['data-timezone_offset' => $timezoneOffsetSeconds]
+                            'attr' => ['data-timezone-offset' => $timezoneOffsetSeconds]
                         ]
                     )
                     ->setAutoInitialize(false)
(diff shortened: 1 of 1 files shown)


lib/FieldType/Mapper/DateFormMapper.php
```diff
@@ -59,6 +60,7 @@ public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
                         [
                             'required' => $fieldDefinition->isRequired,
                             'label' => $fieldDefinition->getName(),
+                            'attr' => ['data-timezone_offset' => $timezoneOffsetSeconds]
```
dew326: data attributes should be with `-` not with `_`. Thanks to that in `dataset` you have the camelCase properties

barw4: Added `-` as requested.

lib/FieldType/Mapper/DateFormMapper.php
```diff
@@ -59,6 +60,7 @@ public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
                         [
                             'required' => $fieldDefinition->isRequired,
                             'label' => $fieldDefinition->getName(),
+                            'attr' => ['data-timezone-offset' => $timezoneOffsetSeconds]
```
adamwojs: This data attribute is specific for date picker component used in `ezplatform-admin-ui` so it should be generated here https://github.com/ezsystems/ezplatform-admin-ui/blob/1.5/src/bundle/Resources/views/fieldtypes/edit/ezdate.html.twig

barw4: Resolve in: https://github.com/ezsystems/ezplatform-admin-ui/pull/1185/commits/32cd1a0d4eaa202b7da88b076cce4951bf20726d

PR #333, opened 2020-02-13
EZP-31310: Fixed editing for content with multiple locations

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31310
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Tests pass?   | yes/no
| Doc needed?   | yes/no
| Related to | <ul><li>https://github.com/ezsystems/ezplatform-admin-ui/pull/1237 </li><li>https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/269</li></ul>

When the user wants to edit the content that has multiple locations and occurs checking if unpublished draft exists then we need to know in which location the user tries to do this. This is needed to prepare proper links to edit versions. We should do this when it is possible (when a user is on the location view). When the user wants to edit content from a dashboard then the location is unknown. Therefore ContentEditViewBuilder must not only check the main Location but also other possible locations to which a user may have access. 

#### Checklist:
- [ ] Coding standards (`$ composer fix-cs`)
- [ ] Ready for Code Review

### DIFF
--- a/lib/Content/View/Builder/ContentEditViewBuilder.php
+++ b/lib/Content/View/Builder/ContentEditViewBuilder.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 
 namespace EzSystems\RepositoryForms\Content\View\Builder;
 
+use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
 use eZ\Publish\API\Repository\Repository;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\API\Repository\Values\Content\Language;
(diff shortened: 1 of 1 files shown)


lib/Content/View/Builder/ContentEditViewBuilder.php
```diff
@@ -8,8 +8,10 @@
 
 namespace EzSystems\RepositoryForms\Content\View\Builder;
 
+use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
 use eZ\Publish\API\Repository\Repository;
 use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
```
andrerom: ?

mikadamczyk: 👍

lserwatka:
Could you merge it up?

PR #334, opened 2020-02-18
EZEE-2930: Added Language obj. instance to User create/edit views

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZEE-2930](https://jira.ez.no/browse/EZEE-2930)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Another PR:
https://github.com/ezsystems/ezplatform-admin-ui/pull/1244

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Ready for Code Review

### DIFF
--- a/bundle/Controller/UserController.php
+++ b/bundle/Controller/UserController.php
@@ -6,6 +6,7 @@
  * @license For full copyright and license information view LICENSE file distributed with this source code.
  * @version //autogentag//
  */
+
 namespace EzSystems\RepositoryFormsBundle\Controller;
 
 use eZ\Bundle\EzPublishCoreBundle\Controller;
@@ -16,7 +17,6 @@ use eZ\Publish\API\Repository\LanguageService;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\PermissionResolver;
 use eZ\Publish\API\Repository\UserService;
-use eZ\Publish\Core\Base\Exceptions\BadStateException;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentType;
 use EzSystems\RepositoryForms\Data\Mapper\UserCreateMapper;
@@ -120,13 +120,15 @@ class UserController extends Controller
             }
         }
 
-        return new UserCreateView(null, [
-            'form' => $form->createView(),
-            'language' => $language,
-            'parentLocation' => $location,
-            'contentType' => $contentType,
-            'parentGroup' => $parentGroup,
-        ]);
+        return new UserCreateView(
+            null, [
+                'form' => $form->createView(),
+                'language' => $language,
+                'parentLocation' => $location,
+                'contentType' => $contentType,
+                'parentGroup' => $parentGroup,
+            ]
+        );
     }
 
     /**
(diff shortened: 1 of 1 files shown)


bundle/Controller/UserController.php
```diff
@@ -155,20 +155,21 @@ public function editAction(
         ?string $language = null,
         Request $request
     ) {
+        $languageCode = $language;
```
adamwojs: You can simply rename `$language` parameter to `$languageCode`, it won't break BC. 

```suggestion
```

bundle/Controller/UserController.php
```diff
@@ -155,20 +155,21 @@ public function editAction(
         ?string $language = null,
```
ViniTou: This could be renamed to `$languageCode` to simplify things.

bundle/Controller/UserController.php
```diff
@@ -181,11 +182,15 @@ public function editAction(
             }
         }
 
+        $location = $this->locationService->loadLocation($user->versionInfo->contentInfo->mainLocationId);
```
mikadamczyk: User could be assigned to multiple locations (user group) and it is possible situation when main location is not accessible here (by permissions). This sound like an edge case but we have 3. line issue https://jira.ez.no/browse/EZP-31310

bundle/Controller/UserController.php
```diff
@@ -155,20 +155,21 @@ public function editAction(
...
         Request $request
     ) {
+        $languageCode = $language;
         $user = $this->userService->loadUser($contentId);
         if (!$this->permissionResolver->canUser('content', 'edit', $user)) {
             throw new CoreUnauthorizedException('content', 'edit', ['userId' => $contentId]);
         }
         $contentType = $this->contentTypeService->loadContentType($user->contentInfo->contentTypeId);
-
+        $language = $this->languageService->loadLanguage($languageCode ?:$this->languageService->getDefaultLanguageCode());
```
adamwojs: Are you sure about `?: $this->languageService->getDefaultLanguageCode()` ?

bundle/Controller/UserController.php
mateuszdebinski: Corrected

bundle/Controller/UserController.php
mateuszdebinski: I changed these parameters to mandatory after talked Maciej, then LanguageCode always must exist and the user must have permission

bundle/Controller/UserController.php
```diff
@@ -181,11 +182,25 @@ public function editAction(
             }
         }
 
+        try {
+            // assume main location if no location was provided
+            $location = $this->locationService->loadLocation((int)
+            $user->versionInfo->contentInfo->mainLocationId);
```
alongosz: This is not very readable. Please use PhpStorm to reformat your code.

bundle/Controller/UserController.php
```diff
@@ -181,11 +182,25 @@ public function editAction(
...
+        } catch (UnauthorizedException $e) {
+            // if no access to the main location assume content has multiple locations and first of them can be used
+            $availableLocations = $this->locationService->loadLocations($user->versionInfo->contentInfo);
+            $location = array_shift($availableLocations);
```
alongosz: I'm not entirely sure about that. AFAIK that Controller is used in Location view, so switching seamlessly to another location doesn't seem correct.

bundle/Controller/UserController.php
```diff
@@ -181,11 +182,25 @@ public function editAction(
...
+
+
         return new UserUpdateView(null, [
             'form' => $form->createView(),
             'languageCode' => $language,
+            'language' => $this->languageService->loadLanguage($language),
             'contentType' => $contentType,
             'user' => $user,
+            'location' => $location,
+            'parentLocation' => $location
```
alongosz: Location and parent Location cannot be the same

lserwatka:
Please merge it up

PR #332, opened 2020-01-18
EZP-31317: Loaded ContentType using preferred language in UserController

alongosz:
|     |     |
| --- | --- |
| **JIRA issue** | [EZP-31317](https://jira.ez.no/browse/EZP-31317) |
| **Type** | Bug
| **Target version** | 2.5

User Content create & edit actions need to load Content Type in current user's preferred language so Field Definition names used to display labels in create & edit forms are displayed in a proper language.
For full description and steps to reproduce please refer to [EZP-31317](https://jira.ez.no/browse/EZP-31317).

Merge up: since in eZ Platform 3.0 we use SA-aware layer, I expect this issue does not exist there, but we need to check that (not sure if SA-aware layer has user preferred language set).

**TODO**:
- [ ] Provide behat coverage (as a follow-up?)

PR #337, opened 2020-04-07
[Travis] Fixed code style check

alongosz:
Seems like for `master` there was an intent to run a CS job, but the script actually didn't do anything except installing dependencies. Moved `install` and `script` to proper job, so it's more readable.

:exclamation: The PR includes a merge-up of #336 and #335 , so needs to be merged fast-forward once Travis is passing.

#335 changes were automatically dropped as they were moved elsewhere I guess.

**TODO**:
- [x] Wait for Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ matrix:
     # 7.3
     - name: "Code Style Check"
       php: 7.3
-      env: CHECK_CS=1
+      install: travis_retry composer install --prefer-dist --no-interaction --no-suggest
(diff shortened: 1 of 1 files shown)


mnocon:
Looks better, thanks!
Question: should be maintain this branch at all? I thought that it has been replaced by content-forms and it's obsolete (I might be wrong)

alongosz:
@mnocon due to git-flow we rather should

alongosz:
Merging to unblock porting of #332 to `ezplatform-content-forms`.

PR #338, opened 2020-04-08
EZP-31449: Adapted TranslatablePropertyTransformer to properly handle nulls

barw4:
| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | [EZP-31449](https://jira.ez.no/browse/EZP-31449)
| **Bug**| yes
| **New feature**    | no
| **BC breaks**      |no
| **Tests pass**     | yes
| **Doc needed**     | no

Reverse transformation in `TranslatablePropertyTransformer` was adapted to properly handle `ContentType` and `ContentType's`  fieldtype translatable fields when the empty value was provided in form.

Original kernel PR and discussion: https://github.com/ezsystems/ezpublish-kernel/pull/2988 -> closed right now.

**TODO**:
- [x] Fix a bug.
- [x] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

### DIFF
--- a/tests/RepositoryForms/Form/DataTransformer/TranslatablePropertyTransformerTest.php
+++ b/tests/RepositoryForms/Form/DataTransformer/TranslatablePropertyTransformerTest.php
@@ -62,8 +62,8 @@ class TranslatablePropertyTransformerTest extends TestCase
     public function reverseTransformProvider()
     {
         return [
-            [false, 'fre-FR', null],
-            [null, 'fre-FR', null],
+            [false, 'fre-FR', ['fre-FR' => false]],
+            [null, 'fre-FR', ['fre-FR' => null]],
             ['français', 'fre-FR', ['fre-FR' => 'français']],
             ['english', 'eng-GB', ['eng-GB' => 'english']],
             ['norsk', 'nor-NO', ['nor-NO' => 'norsk']],
--- a/tests/RepositoryForms/Form/DataTransformer/TranslatablePropertyTransformerTest.php
+++ b/tests/RepositoryForms/Form/DataTransformer/TranslatablePropertyTransformerTest.php
@@ -62,7 +62,7 @@ class TranslatablePropertyTransformerTest extends TestCase
     public function reverseTransformProvider()
     {
         return [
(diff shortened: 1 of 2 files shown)


tests/RepositoryForms/Form/DataTransformer/TranslatablePropertyTransformerTest.php
```diff
@@ -62,8 +62,8 @@ public function testReverseTransform($inputValue, $languageCode, $expected)
     public function reverseTransformProvider()
     {
         return [
-            [false, 'fre-FR', null],
-            [null, 'fre-FR', null],
+            [false, 'fre-FR', ['fre-FR' => false]],
```
alongosz, line 65: not so sure about storing false for false. I would stick to `null` when empty.

barw4:
> Can we narrow it down to just Content Types? Or is this transformer used only there?
> 
> POV ping @webhdx

@alongosz Its also used in CT's fieldtypes and the bug there is exactly the same - so this fixes both of them - it's not used anywhere else.

tests/RepositoryForms/Form/DataTransformer/TranslatablePropertyTransformerTest.php
barw4: @alongosz basically, I do not think there is a point at even testing for "false", there won't be any situation that would output us a "false" there. When it's empty - it's just "null".

ViniTou: Testing for false is good, as those are unit tests and we should test for unexepected values are there is no typehinting there.

I would just say, it should be `[false, 'fre-FR', ['fre-FR' => null]],`

@alongosz

barw4: @ViniTou resolved in https://github.com/ezsystems/repository-forms/pull/338/commits/847c0b6268e4ec76df0b991892ea39b6af3b9569

PR #339, opened 2020-04-15
EZP-31542: Added PHP 7.4 support (eZ Platform 2.5)

adamwojs:
| Question                                  | Answer
| ---------------------------------------- | ------------------
| **JIRA issue**                          | [EZP-31542](https://jira.ez.no/browse/EZP-31542)
| **Type**                                   | improvement
| **BC breaks**                          | no
| **Tests pass**                          | yes
| **Doc needed**                       | no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
-    "name": "ezsystems/repository-forms",
     "description": "Use Symfony forms with eZ Platform repository objects",
+    "name": "ezsystems/repository-forms",
     "license": "GPL-2.0-only",
     "type": "ezplatform-bundle",
     "authors": [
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -16,8 +16,8 @@
         "jms/translation-bundle": "^1.3"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7|^6.0",
-        "matthiasnoback/symfony-dependency-injection-test": "~1.0|~2.0",
+        "phpunit/phpunit": "^5.7|^6.0|^7.5",
```
alongosz, line 19: Does it really make sense to still have this `|`? It's require-dev package, so in a deterministic way only a specific PHPUnit version will be installed and I expect it to be `^7.5`, because it supports all the PHP versions in the test matrix. Keep in mind that `|` causes more memory usage for Composer (1.x)

adamwojs: Fixed in https://github.com/ezsystems/repository-forms/pull/339/commits/02720f8bfd981ebb96d00c227ec1dbd1e97b84e3

PR #348, opened 2021-06-16
IBX-154: Allowed to define if the alternative text for image field is required (2.5)

adamwojs:
> JIRA: [https://issues.ibexa.co/browse/IBX-154](https://issues.ibexa.co/browse/IBX-154)

### DIFF
--- a/lib/Form/Type/FieldType/ImageFieldType.php
+++ b/lib/Form/Type/FieldType/ImageFieldType.php
@@ -8,6 +8,8 @@ namespace EzSystems\RepositoryForms\Form\Type\FieldType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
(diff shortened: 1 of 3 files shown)


lib/Form/Type/FieldType/ImageFieldType.php
```diff
@@ -38,13 +38,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 TextType::class,
                 [
                     'label' => /** @Desc("Alternative text") */ 'content.field_type.ezimage.alternative_text',
-                    'required' => false,
+                    'required' => $options['is_alternative_text_required'],
```
alongosz, line 43: Is it guaranteed that the key is present in the array?

alongosz (approved):

lib/Form/Type/FieldType/ImageFieldType.php
Steveb-p: As part of `configureOptions`, yes.

`setDefaults` implies that the field is required, at the same time providing a default value.

Steveb-p (approved):

ViniTou (approved):

pbisaga (approved):
QA approved, checked on eZ Platform Enterprise 2.5 with .diff

